PR #52, opened 2015-04-06
Optimise and refine

neilj:
This pull request encompasses the following refinements:
1. Optimise: without changing the semantics, many easy optimisations were made, for example changing from searching an array for allowed tags/attributes [O(n)] to constructing a lookup table [O(1)].
2. Fix small documentation errors, typos and whitespace inconsistencies.
3. Add an "isSupported" property to the DOMPurify object to allow users to easily check if the browser supports the full DOMPurify.
4. Fix a bug where sanitize would return a string even when RETURN_DOM is set, if the input had no "<" character.
5. Always return the full html including browser inferred `<html><body>` etc. tags if WHOLE_DOCUMENT config option is set, even if the input has no "<" character.

The test suite has one modification for (5). No other semantic changes were made and the full test suite runs without error in latest Chrome.

### DIFF
--- a/purify.js
+++ b/purify.js
@@ -326,7 +326,6 @@
             ) {
                 /* Keep content for white-listed elements */
                 if (KEEP_CONTENT && currentNode.insertAdjacentHTML
-                    && currentNode.nodeName.toLowerCase
                     && CONTENT_TAGS.indexOf(tagName) !== -1){
                     try {
                         currentNode.insertAdjacentHTML('AfterEnd', currentNode.innerHTML);
--- a/purify.js
+++ b/purify.js
@@ -352,7 +352,7 @@
          * _sanitizeAttributes
          *
          * @protect attributes
-         * @protect removeAttribiuteNode
+         * @protect removeAttributeNode
          * @protect setAttribute
          * @protect cloneNode
          *
--- a/purify.js
+++ b/purify.js
@@ -475,7 +475,7 @@
         cfg ? _parseConfig(cfg) : null;
 
         /* Exit directly if we have nothing to do */
-        if (dirty.indexOf('<') === -1) {
+        if (!RETURN_DOM && !WHOLE_DOCUMENT && dirty.indexOf('<') === -1) {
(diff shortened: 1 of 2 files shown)


mathiasbynens:
Can you add a test for the `RETURN_DOM` change?

neilj:
Sure, no problem. Done.

fhemberger:
Cool, thank you!

mathiasbynens:
Also please add a test that checks if the `isSupported` property is present (so we don’t accidentally remove/rename it in the future). Thanks!

neilj:
Sure, done.

mathiasbynens:
@neilj

cure53:
Thanks, @neilj! That was a lot of work and very good changes. Very much appreciated. 

I am currently reviewing if the changes introduced any security regressions (specifically on MSIE). Once done, I am ready to merge. Should be the case by around tomorrow.

neilj:
Thanks @cure53; code review is definitely important in a project like this! I'm looking into using DOMPurify as an extra layer of defence in the [FastMail](https://www.fastmail.com) web UI, so I was more than happy to spend a few hours this morning tidying it up as I read through and understood the code.

cure53:
@neilj Fully agreed, and if you don't mind, I would like to add your name to the acknowledgements section in the README.md. In case all is fine on MSIE10+ as well (cannot test right now, am on the road), I will probably trigger a new release tomorrow as well. Thanks again :)

neilj:
Sure, that would be great. Thanks.

cure53:
Tests are working well on IE10+, FF and Spartan. I am ready for a merge. The change-review was successful as well - no objections from my side. 

Any objections, @fhemberger and @mathiasbynens?

fhemberger:
LGTM :+1:

mathiasbynens:
Belated LGTM :+1: (https://github.com/cure53/DOMPurify/pull/52#issuecomment-89962686 already implied it, kinda)

PR #56, opened 2015-04-28
Workaround Safari Crash

neilj:
Safari (latest version on both iOS + Mac) will crash hard if you try to remove a "name" attribute from an image node that also has an "id" attribute. I've filed a bug with Apple, but in the meantime this patch works around the issue.

Also, I've fixed up the tests so they all pass in Safari: it has slightly different behaviour parsing form tags.

### DIFF
--- a/purify.js
+++ b/purify.js
@@ -593,7 +593,9 @@
      * @return void
      */
     DOMPurify.removeHook = function(entryPoint) {
-        if (hooks[entryPoint]) {hooks[entryPoint].pop()}
+        if (hooks[entryPoint]) {
+            hooks[entryPoint].pop();
+        }
     };
 
     /**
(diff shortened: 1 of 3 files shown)


mathiasbynens:
Is there a link to the bug report on bugs.webkit.org?

neilj:
I've filed one in radar and spent an hour trying to file one on bugs.webkit.org, but my internet connection was so bad today here in Vietnam (where I am at the moment), that I couldn't successfully submit it (some parts of the internet are OK, but some routes have insanely high packet loss). I will try again soon and put a link here when I succeed!

neilj:
OK, WebKit bug filed: https://bugs.webkit.org/show_bug.cgi?id=144371

PR #57, opened 2015-05-07
Allow more benign contents/attributes

neilj:
This pull request makes the following changes:

1. Preserve contents of <o:p> tag
2. Allow background, cellpadding and cellspacing attributes

This is important for preserving the formatting of many HTML emails. It should not have any security implications.

### DIFF
--- a/purify.js
+++ b/purify.js
@@ -96,16 +96,17 @@
     var DEFAULT_ALLOWED_ATTR = _addToSet({}, [
 
         // HTML
-        'accept','action','align','alt','autocomplete','bgcolor','border',
-        'checked','cite','class','color','cols','colspan','coords','datetime',
-        'default','dir','disabled','download','enctype','for','headers','height',
-        'hidden','high','href','hreflang','id','ismap','label','lang','list',
-        'loop', 'low','max','maxlength','media','method','min','multiple',
-        'name','novalidate','open','optimum','pattern','placeholder','poster',
-        'preload','pubdate','radiogroup','readonly','rel','required','rev',
-        'reversed','rows','rowspan','spellcheck','scope','selected','shape',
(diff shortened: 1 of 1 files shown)


fhemberger:
Uhh …stawp! `<o:p>` is not a valid HTML tag, and as such is not a part of DOMPurify.

Can you please use `var clean = DOMPurify.sanitize(dirty, {ADD_TAGS: ['o:p']});` in this case instead?

neilj:
The tag is not being included. Just the *contents*.

So `<p><o:p>foo</o:p></p>` now is sanitised to `<p>foo</p>` instead of `<p></p>`.

fhemberger:
Ok, I see. Still, I would prefer an option like `ADD_CONTENT_TAGS` for those cases.

neilj:
I see your point. On the other hand, anyone that wants to use this to sanitise the crap that MS Office produces and calls HTML is going to need to add it, and may not find out immediately that they need to add it and have to track down what's causing issues when they get bug reports. Better for things to just work IMO.

fhemberger:
Hmm, are there any special tags produced by LibreOffice/OpenOffice as well?

neilj:
Not sure, haven't tested it. It's not used for composing email so I'm unlikely to come across it. I have a vague memory that its HTML is a lot cleaner, but I honestly haven't used the product in about 10 years.

cure53:
Hi all,

I was hesitant as well first but then again, DOMPurify's goal is not to produce output that conforms any ever-changing standards but to master the split between maximum security and maximum tolerance. Thus I decided that `<o:p>` is a go. It doesn't violate our goals and there fore, if needed, should be considered. 

As for LibreOffice: I think no, I had a closer look at that in the past and didn't find any particular weirdness on the HTML layer of things.

Any thoughts?
.mario

PR #58, opened 2015-05-15
Improve logging for assert.contains

Joris-van-der-Wel:
I am working on fixing failing test cases in jsdom, but the output for many test cases is not very useful. Especially because a large amount of test cases have the same name. This change shows the following info when a case fails:

1. Show the actual output instead of "false"
2. Show a list of all valid expected values
3. Show the payload


cure53:
Oh, sweet, I was gonna do that a long time ago but never got to it. Thanks! Much appreciated!

PR #60, opened 2015-05-17
Support for custom `window`

Joris-van-der-Wel:
This change lets you set up DOMPurify with your own `window` and `document`. It is now super easy to set up with [jsdom](https://www.npmjs.com/package/jsdom):

```javascript
var document = require('jsdom').jsdom();
var DOMPurify = require('dompurify')(document.defaultView);
console.log(DOMPurify.sanitize('<script>alert("hi");</script><p>hello</p>'));
```

For the extra paranoid, you can put jsdom in a mode wherein it never download or execute scripts:
```javascript
var document = require('jsdom').jsdom('', {
    FetchExternalResources: false,
    ProcessExternalResources: false
});
```

This change is backwards compatible. If you run it in a browser, a `DOMPurify` object is added to the `window`. If you use browserify the following should work exactly like it did before: `require('dompurify').sanitize('...')`. I have never used AMD, so could someone double check that it still works as before?

Another fun thing you can do is browserify jsdom. This might be useful if there is some kind of strange browser that manages to run jsdom, but not DOMPurify (which I doubt, since jsdom uses ES6).

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,5 +1,5 @@
 {
-  "browser": true,
+  "browser": false,
   "curly": true,
   "eqeqeq": true,
   "expr": true,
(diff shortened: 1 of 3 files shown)


cure53:
Thanks!

Issue #29, opened 2014-07-06
Make DOMPurify work in Node.js


fhemberger:
Regarding issue #26 and #27, I originally held back Common JS style exports and publishing on npm on purpose, as DOMPurify doesn't run on a pure Node.js environment (it does client side with Browserify).

I'm still looking for a way to get it to work on Node.js as well. jsdom [lacks DOM Level 2 Traversal methods](https://github.com/tmpvar/jsdom/issues/317) like `createNodeIterator` at the moment, which DOMPurify uses internally.

What's currently missing:
- [x] `document.implementation.createHTMLDocument`  
  (polyfilled with `return jsdom('<html><body></body></html>');`)
- [x] `window.NodeFilter (extracted the properties needed for DOMPurify from the spec)
- [ ] `document.createNodeIterator(root, whatToShow, filter, entityReferenceExpansion)`
- [ ] `NodeIterator.nextNode()` implementation
- [ ] Setter method for `document.body.outerHTML`

wtfuii:
+1 for this feature..

Joris-van-der-Wel:
jsdom 5.1.0 supports NodeIterator: https://github.com/tmpvar/jsdom/pull/1092
There are still other issues though, such as `firstElementChild`

fhemberger:
Thanks, I read the announcement (but hadn't notice the version has already been released).
I hope to find the time soon to look into this topic again. Sorry this takes so long, but fiddling around with a DOM implementation in Node is really messy. ;)

Joris-van-der-Wel:
[jsdom 5.2.0](https://github.com/tmpvar/jsdom/blob/master/Changelog.md#520) has many fixes to make dompurify work. Most html snippets can now be "purified" properly.
Although I have to load dompurify as a script using jsdom.env because dompurify uses a lot of globals. It would be great if I could instantiate dompurify by passing it a `Window`.

There are still a few test cases failing though, which would require more work in jsdom. One example is that jsdom currently throws an exception if an attribute is set with an invalid name: `foo.innerHTML = "<img id/=' >"`. The spec says that resolving parser errors is optional, but it is of course not how the browsers behave. The result is that dompurify is not able to clean up such snippets.

Joris-van-der-Wel:
Above has been fixed. There are now 13 failing test cases.
Additional stuff to fix in jsdom:

* A few test cases with svg tags fail, the output looks safe, but it is not what the test cases expect
* stub or implement adoptNode
[...]

PR #64, opened 2015-05-29
Update _isClobbered to check currently required properties

neilj:
The _isClobbered fn detects whether any of the methods/attributes we need on
the element have been clobbered. The properties it checks for had not been
updated with the changes in how the sanitisation methods manipulated the
elements, meaning it was making some unnecessary checks, and not making some
required ones, which could result in an error being thrown with certain input.

### DIFF
--- a/purify.js
+++ b/purify.js
@@ -34,11 +34,10 @@
     var HTMLHtmlElement = window.HTMLHtmlElement;
     var DocumentFragment = window.DocumentFragment;
     var HTMLBodyElement = window.HTMLBodyElement;
-    var HTMLCollection = window.HTMLCollection;
     var HTMLTemplateElement = window.HTMLTemplateElement;
     var NodeFilter = window.NodeFilter;
-    var NodeList = window.NodeList;
     var Text = window.Text;
+    var toString = Object.prototype.toString;
 
     var hooks = {};
 
@@ -316,24 +315,18 @@
             return false;
         }
         if (
-            (elm.children && !(elm.children instanceof HTMLCollection))
-            || (elm.attributes instanceof HTMLCollection)
-            || (elm.attributes instanceof NodeList)
+            (elm.outerHTML && typeof elm.outerHTML !== 'string')
             || (elm.insertAdjacentHTML && typeof elm.insertAdjacentHTML !== 'function')
-            || (elm.outerHTML && typeof elm.outerHTML !== 'string')
             || typeof elm.nodeName !== 'string'
             || typeof elm.textContent !== 'string'
-            || typeof elm.nodeType !== 'number'
-            || typeof elm.COMMENT_NODE !== 'number'
-            || typeof elm.setAttribute !== 'function'
-            || typeof elm.hasAttribute !== 'function'
-            || typeof elm.cloneNode !== 'function'
-            || typeof elm.removeAttributeNode !== 'function'
             || typeof elm.removeChild !== 'function'
-            || typeof elm.attributes.item !== 'function'
-            || (elm.id === 'createElement' || elm.name === 'createElement')
-            || (elm.id === 'implementation' || elm.name === 'implementation')
-            || (elm.id === 'createNodeIterator' || elm.name === 'createNodeIterator')
+            || !elm.attributes
+            || toString.call( elm.attributes ) !== '[object NamedNodeMap]'
+            || typeof elm.removeAttribute !== 'function'
+            || typeof elm.setAttribute !== 'function'
+            || elm.id === 'createElement' || elm.name === 'createElement'
+            || elm.id === 'implementation' || elm.name === 'implementation'
+            || elm.id === 'createNodeIterator' || elm.name === 'createNodeIterator'
         ) {
             return true;
         }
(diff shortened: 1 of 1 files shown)


Joris-van-der-Wel:
This breaks in jsdom (although this is jsdom's fault) https://github.com/tmpvar/jsdom/issues/1136 #29

```javascript
doc.createElement('div').attributes.toString() // '[object Object]' in jsdom`
```

Joris-van-der-Wel:
Actually, this is not fixable in jsdom until new ES6 features land in iojs

cure53:
Ew, I see. You would requite `Symbol.toStringTag()` I assume? 
Can we work around that or should we revert?

neilj:
We could just change it to just use `attributes.toString()`, then it's fixable without ES6. I don't *think* this is a security risk (someone please double check you can't clobber to toString property of the clobbered attributes property).

This is then easily shimmable.

I've done this in commit 486f155.

Joris-van-der-Wel:
Using instanceof should also work:

```javascript
var document = require('jsdom').jsdom()
var window = document.defaultView;

var purifyDoc = document.implementation.createHTMLDocument();
purifyDoc.createElement('div').attributes instanceof window.NamedNodeMap === true
```

Joris-van-der-Wel:
Yes, it would require `attributes[Symbol.toStringTag] = ...`

neilj:
Oh, good call. Yes, that's a much better solution.

cure53:
I think the `instanceof` approach is a bit risky. I am currently testing with this example:

```html
<form onclick=alert(attributes)>
clickme (expecting somthing else)
<input name=attributes>
<input name=attributes>
</form>

<form onclick=alert(attributes)>
clickme (expecting a NamedNodeMap)
<input name=bla>
<input name=blubb>
</form>
```

In current Gecko and Blink we get a `[object RadioNodeList]`, in MSIE11 and 10 we get `[object HTMLCollection]` and I am not sure if it's really impossible to clobber it to become a named node map. We had one bypass like that in the past - so I am a bit anxious here ;)

neilj:
I'm fairly sure it's not possible to clobber it into a NamedNodeMap. It's certainly still strictly more secure than what we had before my previous patch that you just landed. The old test was:

    (elm.attributes instanceof HTMLCollection) ||
    (elm.attributes instanceof NodeList) ||
    typeof elm.attributes.item !== 'function'

If you could have clobbered it to be a NamedNodeMap, it would have passed all these tests!

I've done pull request #66 if you're happy to land the change.

cure53:
LGTM!

Joris-van-der-Wel:
Also, it appears that in firefox 22 it was named `MozNamedAttrMap`. So this would affect both the instanceof and toString() solutions.

neilj:
Yeh, just reading that: it was changed back in FF34. Let's get the constructor directly from the browser: see pull request #67. (Note, I don't have an old FF to test with and my internet connection at the moment is too slow to download it, so I'd appreciate someone double checking this works in all browsers first! It certainly works in the current Chrome/Safari/FF/Opera and Opera 12).

Joris-van-der-Wel:
Hah. should set up testem + saucie or (something similar) for this kind of stuff.

neilj:
Bah, no that doesn't work (see the comment in #67). If someone can download FF22, I have another idea that should work:

    var NamedNodeMap = window.NamedNodeMap || window.MozNamedAttrMap;

I'm not sure if NamedNodeMap is still defined in FF22 though (even though it's not used for attributes). If it is, we need to switch the order:
    
    var NamedNodeMap = window.MozNamedAttrMap || window.NamedNodeMap;

I think that should work.

Joris-van-der-Wel:
Here's another questionable solution:
```
Object.getPrototypeOf(document.body.attributes) === Object.getPrototypeOf(document.createElement('form').attributes)
```

cure53:
I think I prefer @neilj's approach here. We can clobber the prefixed version though in newer FF (and all other browsers):

```html
<form id="MozNamedNodeMap"></form>
<script>alert(MozNamedNodeMap)</script>
```

Joris-van-der-Wel:
I tried this in Firefox 22 Linux:

```html
<!DOCTYPE html>
<html>
	<head><title>DOMPurify temporary stuff</title></head>
	<body>
		<p>foo</p>
		<p><script>var NamedNodeMap = window.NamedNodeMap || window.MozNamedAttrMap;
		document.write(document.body.attributes.toString());
		document.write(document.body.attributes instanceof NamedNodeMap);
		</script></p>
		<p><script>document.write(Object.getPrototypeOf(document.body.attributes) === 
				Object.getPrototypeOf(document.createElement('form').attributes)
		);</script></p>
		<p>bar</p>
	</body>
</html>

```

Result: https://saucelabs.com/jobs/48fbfec372704992922c8f3fc7ff71e5/0002screenshot.png

neilj:
@cure53 I don't think the clobbering of NamedNodeMap or MozNamedAttrMap matters, because we cache our reference to it when purify.js is first evaluated, (presumably) before we pass in any malicious content, so before any clobbering can take place. The clobbering will overwrite the global var, but we already have a local var with a reference.

@Joris-van-der-Wel I'm curious to double check: is window.NamedNodeMap defined in FF22? If so, we don't need to change anything (because it's already supporting backwards compatibility). Otherwise, it looks like my patch will work fine.

cure53:
@neilj You are right!

Joris-van-der-Wel:
@neilj It is __not__ defined. Also its typeof MozNamedAttrMap is "object", while typeof "NamedNodeMap" is "function"

neilj:
OK, so this patch is good:

    var NamedNodeMap = window.NamedNodeMap || window.MozNamedAttrMap;

Made pull request #70.

PR #66, opened 2015-05-29
Use instanceof instead of toString to check if attributes is clobbered

neilj:
This makes it much easier to implement compatible support in JS DOM
emulation libraries.


cure53:
Thanks :)

PR #67, opened 2015-05-29
Get constructor for Attributes from browser.

neilj:
FF22-34 used a different type for the attributes collection. Get it directly
from the browser to ensure we have it right.


neilj:
No, scrap that. In Safari/Chrome, the constructor is also a constructor for elements (it doesn't return NamedNodeMap for some reason)! So this very much fails in its intended purpose.

PR #68, opened 2015-05-29
RETURN_DOM_FRAGMENT and RETURN_DOM_IMPORT

### LINKED ISSUE #62: Option to return a DocumentFragment instead of html
The way I am currently using dompurify looks like this:

```javascript
var html = this.databaseEntity.foo;
html = dompurify(this.document.defaultView).sanitize(html);
var fragment = parseHTMLSnippit(this.document.defaultView, html);
this.bar.appendChild(fragment);
```

Which means the html is being parsed twice. It would be useful if there was an option in dompurify to return a DocumentFragment. (adopted/imported into the document dompurify has been constructed with). 

For example:

```javascript
var html = this.databaseEntity.foo;
var fragment = dompurify(this.document.defaultView).sanitize(html, {output: 'fragment'});
this.bar.appendChild(fragment);
```

Joris-van-der-Wel:
Fixes #62

Should work properly even if `SANITIZE_DOM` is disabled.

```javascript
document.body.appendChild(
    DOMPurify.sanitize(dirty, {RETURN_DOM_FRAGMENT: true, RETURN_DOM_IMPORT: true})
);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,8 +99,11 @@ var clean = DOMPurify.sanitize(dirty, {ADD_ATTR: ['my-attr']});
 // prohibit HTML5 data attributes (default is true)
 var clean = DOMPurify.sanitize(dirty, {ALLOW_DATA_ATTR: false});
 
-// return a DOM instead of an HTML string (default is false)
+// return a DOM HTMLBodyElement instead of an HTML string (default is false)
 var clean = DOMPurify.sanitize(dirty, {RETURN_DOM: true});
+
+// return a DOM DocumentFragment instead of an HTML string (default is false)
+var clean = DOMPurify.sanitize(dirty, {RETURN_DOM_FRAGMENT: true});
 
 // return entire document including <html> tags (default is false)
 var clean = DOMPurify.sanitize(dirty, {WHOLE_DOCUMENT: true});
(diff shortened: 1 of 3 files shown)


cure53:
Looks good to me, just checked the commits. 

@fhemberger @neilj What do you think?

Joris-van-der-Wel:
> OK. I can't see how the past names map might be used to compromise security at the moment, but I agree it's better just to use importNode and not worry about it. Might be nice to have a comment in the source code explaining why it was chosen; helps people coming in later (and indeed ourselves in a few months…).

> OK for the options too; I take your point on consistency.

I have added a comment https://github.com/Joris-van-der-Wel/DOMPurify/commit/5618f47feb36766e83a4bab4a8f3388ca1fab463#diff-9d220e0bdde67c89de82589b198242ccR630

Joris-van-der-Wel:
Rebased

cure53:
I am happy with the code and ready for a merge. Any objections for now?

neilj:
Looks good to me with the latest changes. :+1: 

(Sorry for the delay; GitHub didn't email me with the latest comment for some reason).

cure53:
Thank you and no worries :)

PR #69, opened 2015-05-29
add 'offset' attribute for gradients of SVG elements

dhardtke:
I need this for <linearGradients> inside my <svg> tag.

Apparently, DOMPurify removes the 'offset' attribute.


cure53:
LGTM!

PR #72, opened 2015-06-02
Ensure document has blank web-components registry

### LINKED ISSUE #47: Should use template instead of document.implementation.createHTMLDocument to avoid the web-component registry of the owner doc being reused
_initDocument uses document.implementation.createHTMLDocument() to create a new data document.  Because of http://w3c.github.io/webcomponents/spec/custom/#creating-and-passing-registries this is potentially not as safe as it used to be.  Specifically, the newly created document will use the same web-components registry as its execution context, which means that if a dirty HTML string is passed in, a nefarious string could potentially leverage any registered web-components to nefarious ends.

Probably the best course of action is to do something like:
```js
var doc = document.createElement('template').content.ownerDocument;
```
instead of the current document.implementation.createHTMLDocument().  (Note that since it's its own document, I don't think the template node would need to be linked into the DOM, but if something seems wrong, that could be it.  At least in Gecko, things like CSS parsing won't usefully happen until a DOM node is parented into the tree.)

Greater context: We got burnt by this in a non-security context when doing cloneNode, fix part of a bug at https://bugzilla.mozilla.org/show_bug.cgi?id=1116087#c12.

neilj:
Fixes #47 for real, I think. Unless I've misunderstood it, in which case please feel free to reject!

Also implements a better fix for clobbering of required fns on the dirty document; we cache direct references to the fns we need, so it doesn't matter if things get clobbered. This makes the SANITIZE_DOM option more consistent; the behaviour is the same for any name/id that would clobber a method/property in the document/window, regardless of whether we need it internally or not.

### DIFF
--- a/purify.js
+++ b/purify.js
@@ -31,17 +31,25 @@
     }
 
     var document = window.document;
-    var documentImplementation = document.implementation;
-    var createElement = document.createElement;
-    var createNodeIterator = document.createNodeIterator;
-    var importNode = document.importNode;
-    var HTMLHtmlElement = window.HTMLHtmlElement;
     var DocumentFragment = window.DocumentFragment;
-    var HTMLBodyElement = window.HTMLBodyElement;
     var HTMLTemplateElement = window.HTMLTemplateElement;
     var NodeFilter = window.NodeFilter;
     var NamedNodeMap = window.NamedNodeMap || window.MozNamedAttrMap;
     var Text = window.Text;
+
+    // As per issue #47, the web-components registry is inherited by a
+    // new document created via createHTMLDocument. As per the spec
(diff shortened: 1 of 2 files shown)


cure53:
Pinging the original finder, @asutherland. Does that look okay to you as well?

asutherland:
This makes sense to me, with the caveat that I don't really understand the use-case/threat model for the clobbering handling and its tests.  (Not that I think I need to, I just want to make sure it's clear since in the original issue I was only focusing on getting a document free from web-components which may have contributed to this pull request needing to exist.  Apologies about that.)

Specifically, I agree that the creation of the template-document only if the body/html elements were themselves "subclassed" was insufficient to protect against custom element-related issues.  This PR correctly makes it happen and simplifies the control flow path which is also nice.

cure53:
Thank you :)

PR #73, opened 2015-06-03
Fix import node + a variety of minor issues

neilj:
The main fix is for the importNode call, which I inadvertently broke when I fixed the clean document creation issue (#47). Also fixes comment nodes incorrectly being returned as clobbered, removes some more unnecessary clobber checks and does a better check for an id needed to workaround the Safari crash bug.


cure53:
Thanks, LGTM!

PR #75, opened 2015-06-08
fix stupid typo

dhardtke:
how did this get in the .js file? oO

This enables us to use font-weight as an attribute in SVGs.


fhemberger:
Oh, good catch, thanks!

PR #85, opened 2015-07-23
Accept non-alphanumeric data-* attributes

gibson042:
HTML5 allows a broad range of characters for [data-* attribute names](https://html.spec.whatwg.org/multipage/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes), but DOMPurify [further limits them](https://github.com/cure53/DOMPurify/blob/7a031e3232642acd8bbc952015d8c4c229cb40db/purify.js#L479) to alphanumerics, underscores, and hypens. This PR relaxes those and further relies on `setAttribute` to catch truly invalid names, and additionally avoids regex redefinition.

### DIFF
--- a/purify.js
+++ b/purify.js
@@ -476,7 +476,10 @@
                 /* Check the name is permitted */
                 (
                  (ALLOWED_ATTR[lcName] && !FORBID_ATTR[lcName]) ||
-                 (ALLOW_DATA_ATTR && /^data-[\w-]+/.test(lcName))
+                 /* Allow potentially valid data-* attributes
+                    * At least one character after "-" (https://html.spec.whatwg.org/multipage/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes)
+                    * XML-compatible (https://html.spec.whatwg.org/multipage/infrastructure.html#xml-compatible and http://www.w3.org/TR/xml/#d0e804) */
+                 (ALLOW_DATA_ATTR && /^data-[\w.\u00B7-\uFFFF-]/.test(lcName))
                 ) &&
                 /* Get rid of script and data URIs */
                 (
@@ -486,7 +489,7 @@
                   currentNode.nodeName === 'IMG')
                 )
             ) {
-                /* Handle invalid data attribute set by try-catching it */
+                /* Handle invalid data-* attribute set by try-catching it */
                 try {
                     currentNode.setAttribute(name, value);
                 } catch (e) {}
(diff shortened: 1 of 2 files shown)


cure53:
I think the pull request is okay - but I am a bit scared by the very permissive range that is specified now. Now we can even use Unicode control characters like the RTL override as attribute names:

```html
<p data-a‮b="hello">
```

@fhemberger and @neilj what do you think?

fhemberger:
@cure53 Haha, you're the expert on figuring out crazy attack scenarios. ;)

Personally, I haven't come across a site using unicode in data attributes, so I don't know if this is really a big issue at all (even though the standard permits it). Maybe this should be exposed as configuration option (`DATA_ATTR`), I have mixed feelings about putting it into core directly.

I think DOMPurify should be as restrictive/secure as possible at it's core. If you need more permissive settings, you should be able to set them yourself in the configuration (and know exactly what you're doing).

gibson042:
`ALLOW_DATA_ATTR` already exists; it's just that the corresponding regular expression is too restrictive. DOMPurify's biggest gap is not allowing `.`, but [HTML5](https://html.spec.whatwg.org/multipage/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes) does defer to XML.

The range here is wide open because I figured `setAttribute` would catch problems (in fact, I found _it_ too be too restrictive as well—even though e.g. Chromium handles non-BMP `el.innerHTML = '<a data-𝌆="foo"></a>'`, it throws on `el.setAttribute("data-𝌆", "foo")`), but the regex could easily be extended to match [http://www.w3.org/TR/xml/#d0e804](http://www.w3.org/TR/xml/#d0e804) exactly.

gibson042:
```html
<p data-a‮b="hello">
```

I can hardly believe it, but that actually works sanely with `innerHTML` and `getAttribute`, though `setAttribute` fails with the comical
```
'data-a‮b' is not a valid attribute name.
```
and an even more hilarious `DOMException#stack` (one word: "suomynona").

cure53:
Haha, yes. 

I don't see any injections risks right now - so I believe we can accept this PR without any security implications. We are close to publish 0.6.5 and I will have a closer look before that anyway. Thanks!

neilj:
Seems fine from a security standpoint. Code-style wise, I don't really like the regexp declarations below the function in which they're accessed. I'd move them above the sanitizeAttribute function instead, so they're declared before they are first used; it's just a bit clearer if you're reading through the code.

cure53:
Anyone against moving them here?

https://github.com/cure53/DOMPurify/blob/master/purify.js#L227

fhemberger:
:+1:

PR #86, opened 2015-07-23
Fix form aliasing crash

gibson042:
Stripping out `<form>` elements with clobbered `parentNode` crashes DOMPurify.

### DIFF
--- a/purify.js
+++ b/purify.js
@@ -283,6 +283,19 @@
     };
 
    /**
+     * _forceRemove
+     *
+     * @param  a DOM node
+     */
+    var _forceRemove = function(node) {
+        try {
+            node.parentNode.removeChild(node);
+        } catch (e) {
+            node.outerHTML = '';
+        }
+    };
+
+   /**
      * _initDocument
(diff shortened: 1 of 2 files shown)


cure53:
Hmmm, what browser is the crash happening in? DOMPurify itself work well with `parentNode` clobbering and seems to do what it's expected to do.

gibson042:
Confirmed in Chromium and Firefox latest. Did you run the 6eea34c76068a239d626ef58d0b01f9e5f1a62ab test suite? The problem is that the [`ALLOWED_TAGS`/`FORBID_TAGS` element removal on master](https://github.com/cure53/DOMPurify/blob/71f4e2167cc0d9b3040fb42c9d6bb272de14d0c2/purify.js#L386) doesn't include the `outerHTML = ''` fallback that protects `_sanitizeElements`, so `currentNode.parentNode.removeChild(currentNode)` can throw uncaught exceptions.

cure53:
Got it! Thanks :)

PR #99, opened 2015-09-17
Fix using purify.js from /src folder in demos and index.tpl.

### LINKED ISSUE #97: purify.js was moved into src, but the package.json wasn't updated
https://github.com/cure53/DOMPurify/blob/master/package.json#L27

tdeekens:
This replaces all occurrences of `purify.js` to `src/purify.js` and therefore closes #97. Along furthermore removes all trailing whitespaces from demo files.


cure53:
Thanks! Merging...

PR #103, opened 2015-09-21
Oh seven

neilj:
A few small fixes and tidy.
* When the code to cache `document.implementation` was added back, it was in the wrong place so used the wrong `document`.
* Fix return type doesn't respect config if empty input string.
* Revert isSupported check since we fallback still for browsers without DOMParser support.


cure53:
Looking good, thank you :)

```
SUMMARY:

✔ 7184 tests completed

The command "npm test" exited with 0.
```

PR #110, opened 2015-09-29
Fix pull requests not building (breaks tests).

tdeekens:


tdeekens:
...as mentioned tests here fail cause it's a pull requests and you'd have to "force merge".

PR #123, opened 2015-10-25
Adding an example on how to process SVGs

oparoz:
@cure53


cure53:
Nice, thanks! I will review later today.

Issue #131, opened 2015-12-10
src attributes that start with "data:" get stripped out for audio and source tags


styu:
I noticed that you whitelist these src attributes for images - would it be possible to do so for audio and source tags as well?

cure53:
Yep, that should be possible. I don't see any risks connected to that right now - and the change would be minimal. Mind sending a PR for review?

styu:
will do!

cure53:
Thx!

cure53:
I'll prettify and add a test case tomorrow.

cure53:
Build is green, tests all passing, I think we are done here :)

PR #132, opened 2015-12-11
Allow data URIs for audio, video, and source tags in addition to img tags

styu:
https://github.com/cure53/DOMPurify/issues/131


styu:
Hrm is there something extra I need to do?

PR #138, opened 2016-02-14
Improve code quality and robustness

0xsobky:
Hey, guys!

After taking an overview at DOMPurify, I have a few concerns. This pull request tries to address them as explained below:

At line no. 393 and its following line, you're clearly making use of the `m` modifier within these couple regular expressions: `/\{\{.*|.*\}\}/gm` and `/<%.*|.*%>/gm`; that, in fact, has absolutely no effect on the dot meta-character in ECMAScript! I guess you meant to match new lines with that (or line terminators `[\n\r\u2028\u2029]` in general), but the right way to do so is by using a character class as such `[\s\S]`.

Another concern is, at line no. 470, you're making use of a blacklist regex to detect evil URI schemes like `javascript:`, and you are doing it well (albeit, this has been bypassed before). But given that it will only match at the very beginning of a given string, I really wonder what would happen if some random browser doesn't mind evaluating something like `feed:javascript:alert(0)`? This has happened before with Firefox; no guarantee it wouldn't happen again! Not to mention other provisional URI schemes that might extend our attack surface here!

So, my proposed fix for that is to alternatively use this whitelist regex `/^(?:[\W\d]|(?=([a-z]+))\1(?!:)|(?:mailto|tel|(?:http|ftp)s?):)/i` instead--allowing only relative URIs and known good URI schemes like `mailto`, `http`, `ftps`, etc.

Furthermore, at line no. 653, DOMPurify invokes the `toString` method directly on a user-supplied argument (i.e. the `dirty` parameter). I don't think this is a good idea, as such methods are prone to be overridden hence introducing unexpected behaviors; we shouldn't generally be calling methods on arbitrary user-supplied objects without any kind of checks in place.

My proposed fix here is to first check whether our object argument has an overridden `toString` property or not; if so, we borrow a native `toString` method from the right prototype instead...implementing it accordingly for `String` instances (`String.prototype.toString.call(dirty)`), arrays (`Array.toString.call(dirty)`), and object literals (`Object.prototype.toString.call(dirty)`). Otherwise, we just invoke the `toString` method directly on the object, as it used to be.

Other than that, just a few convention corrections here and there!

Cheers (and thanks for creating DOMPurify!),

### DIFF
--- a/src/purify.js
+++ b/src/purify.js
@@ -652,7 +652,7 @@
         /* Stringify, in case dirty is an object */
         if (typeof dirty !== 'string') {
             if (typeof dirty.toString !== 'function') {
-                throw TypeError('toString is not a function');
+                throw new TypeError('toString is not a function');
             } else {
                 dirty = dirty.toString();
             }
--- a/src/purify.js
+++ b/src/purify.js
@@ -483,12 +483,11 @@
      * @return  void
      */
     var _sanitizeAttributes = function(currentNode) {
-        /* Execute a hook if present */
-        var hookEvent, l, attr, name, value, lcName, idAttr;
-        var attributes = currentNode.attributes;
-
+        var attr, name, value, lcName, idAttr, attributes, hookEvent, l;
         /* Execute a hook if present */
         _executeHook('beforeSanitizeAttributes', currentNode, null);
+
+        attributes = currentNode.attributes;
 
         /* Check if we have attributes; if not we might have a text node */
         if (!attributes) { return; }
@@ -499,6 +498,7 @@
             keepAttr: true
         };
         l = attributes.length;
+
         /* Go backwards over all attributes; safely remove bad ones */
         while (l--) {
             attr = attributes[l];
@@ -668,7 +668,7 @@
 
         /* Check we can run. Otherwise fall back or ignore */
         if (!DOMPurify.isSupported) {
-            if (typeof window.toStaticHTML === 'object'
+            if (typeof window.toStaticHTML === 'object' 
                 || typeof window.toStaticHTML === 'function') {
                 return window.toStaticHTML(dirty);
             }
--- a/src/purify.js
+++ b/src/purify.js
@@ -633,6 +633,13 @@
         });
     };
 
+    /**
+     * sanitize
+     * Public method providing core sanitation functionality
+     *
+     * @param {String} dirty string
+     * @param {Object} configuration object
+     */
     DOMPurify.sanitize = function(dirty, cfg) {
         var body, currentNode, oldNode, nodeIterator, returnNode;
         /* Make sure we have a string to sanitize.
--- a/src/purify.js
+++ b/src/purify.js
@@ -467,7 +467,7 @@
     };
 
     var DATA_ATTR = /^data-[\w.\u00B7-\uFFFF-]/;
-    var IS_ALLOWED_URI = /^(?:[\W\d]|(?:mailto|tel|(?:http|ftp)s?):|(?=([a-z]+))\1(?!:))/i;
+    var IS_ALLOWED_URI = /^(?:[^a-z]|(?:(?:f|ht)tps?|mailto|tel):|[a-z]+(?:[^a-z:]|$))/i;
     /* This needs to be extensive thanks to Webkit/Blink's behavior */
     var ATTR_WHITESPACE = /[\x00-\x20\xA0\u1680\u180E\u2000-\u2029\u205f\u3000]/g;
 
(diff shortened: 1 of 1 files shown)


cure53:
Heya, thanks for the PR, reviewing now.

cure53:
It looks okay from my side, although I believe, we should add test cases for the new URI handler white-list to make sure nothing gets messed up.

@neilj What do you think?

neilj:
The tidying looks good, as does the new regex for cleaning templates. However, I have a few queries/suggestions regarding the whitelist regex for URL protocols and the toString change:

1. The protocol whitelist regex is composed of 3 alternates. My comments on each:
   * I think `[\W\d]` is more clearly expressed as `[^a-z]`. The latter also matches an opening underscore, but I think this is fine: the point of this part is to say that URLs that don't starting with a letter are safe, and this is more clearly conveyed by this form.
   * `(?:mailto|tel|(?:http|ftp)s?):` is fine, although I wonder whether in the future we might want to allow the list of allowed protocols to be configurable. I can see some apps needing to allow other specific protocols through, while others may wish to prevent some of the above (`tel:` in particular). You could write this part marginally more compactly as `(?:mailto|tel|(?:ht|f)tps?):`, but it's a little less readable.
  * I don't like the use of the lookahead (`?=`) plus back reference. This is liable to be slower (especially as it's capturing), and is hard for anyone reading it to understand. I suggest replacing it with the following, which does the same but should be faster and is much easier to read: `[a-z]+(?:[^a-z:]|$)` – if it starts a-z, the first non-a-z character must not be a `:`.
   * I would then reorder the alternate parts to put the absolute URLs first, as these are most common and the other two parts should be understood together.

So, final form I would recommend:

    /^(?:(?:(?:f|ht)tps?|mailto|tel):|[^a-z]|[a-z]+(?:[^a-z:]|$))/i

2. The current implementation of the `toString` change only avoids calling the `toString` method on the object directly if the property has been redefined on the object itself rather than its prototype. I am not 100% sure exactly what you are trying to guard against here, however I disagree with the toString change in principal – the whole point of having an overridable toString method in JavaScript is so you can implement custom behaviour for this, and stating in the docs that if a non-string is passed it will be converted with a standard `toString()` call rather than the convoluted change proposed is much easier for devs to understand. If we think this is likely to trip people up, a far better solution in my opinion is to throw an error for a non-string, so the developer is forced to convert their object to a string before passing it to DOMPurify. This is reasonable, given the whole purpose of the function, and avoids the dev having to try to understand what the library will do with a non-string.

0xsobky:
@neilj: Good points, however, I think that URIs that start with a slash or a leading double-slash are relatively more common than absolute URLs. Hence I reordered the regex to become `/^(?:[^a-z]|(?:(?:f|ht)tps?|mailto|tel):|[a-z]+(?:[^a-z:]|$))/i`.

And as for the current `toString` implementation, I re-implemented it so it throws a `TypeError` in case of `toString` not being a function.

neilj:
> I think that URIs that start with a slash or a leading double-slash are
> relatively more common than absolute URLs

That would depend on what you're normally sanitising; in my case it's emails, in which *all* links are external. I would contend that the sort of HTML requiring sanitisation is in general more likely to link externally, but I'm not too fussed; the ordering of the regex parts is not super important.

So with the latest changes, I'm happy to give it a thumbs up for merging. Thanks! :+1:

cure53:
Thanks everyone! I merged the PR and fixed the tests, the CI should be green as well soon. I added some minor changes to the URI-safe white-list of attributes to make sure, the new regex doesn't strip out too much.

Looking good now, thanks again, @0xsobky and @neilj!

filedescriptor:
You are right about the ```m``` modifier:
> multiline; treat beginning and end characters (^ and $) as working over multiple lines

as we are not matching beginning or end, I think we can just drop it for simplicity.

In addition, for matching arbitrary character, ```[^]``` can be used to replace ```[\s\S]``` and is perfectly functional (spec-wise and browser-wise).

filedescriptor:
As per [spec](http://tools.ietf.org/html/rfc3986#section-3.1):
> scheme      = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )

You may also consider treating ```[+.-]``` as part of the scheme, especially for ```-``` (```view-source:```, ```moz-icon:```, etc)

0xsobky:
@filedescriptor: While the character set `[^]` is a simpler equivalent to `[\s\S]`, the latter is better supported on older versions of IE (IE9 for instance; not sure if this makes it any better though, as DOMPurify seems to target IE10+) and is arguably clearer for people who are familiar with other regex flavors, hence I originally preferred it over the former.

As for the URI scheme whitelist regex, here's a more comprehensive version (the lookahead with a backreference is necessary here, I think): 
`/^(?:[^a-z]|(?=([a-z+.-]+))\1(?!:)|(?:mailto|tel|(?:ht|f)tps?):)/i`

So, what do you guys think?

cure53:
@0xsobky @filedescriptor I would recommend to keep it with `\s\S` as it is more common and less risky in case the empty negative group is not known by current or future PCRE engines. 

Let me check what the tests say about the new regex!

cure53:
Tests are green, I'd be good to go.

filedescriptor:
I think we generally want to avoid complex regex like this one as it is not pleasure for readability. The logic is also hard to follow, although it may work perfectly fine. (kind of reminds me of those WAF regexes)

Another thing is that with the white-list approach, cases like```<div data-headline="foo:bar"><div>``` are more easier to be removed compared to the original black-list approach. Ideally we should separate the validation between normal attributes and URL attributes.

neilj:
I agree. I would prefer:

    ^(?:(?:(?:f|ht)tps?|mailto|tel):|[^a-z]|[a-z+.\-]+(?:[^a-z+.\-:]|$))

to a complicated reg exp. As before, this is both easier to understand and almost certainly faster.

0xsobky:
@neilj: The order is totally fine, but you can't really replace the lookahead and the backreference here because backtracking! Sadly, it's the only way to implement that comprehensive version of the regex, given that ECMAScript doesn't support lookbehind.

PR #139, opened 2016-02-17
Simplify attribute sanitisation.

neilj:
As per #138, use a simpler regex for IS_ALLOWED_URI. Also cleans up the big `if` condition to make it easier to understand.


filedescriptor:
The new regex looks more expressive now. @0xsobky what do you think?

cure53:
LGTM!

PR #141, opened 2016-02-18
Allow unknown protocols Fixes #140

### LINKED ISSUE #140: v0.7.4 broke <img src=""> handling for us
Hi cure53,

We're using DOMPurify to sanitize emails in the browser, and this means we deal with a fair number of non-standard protocols. Most important is `cid:` to allow inline images stored as attachments (we post-process the email to deal with these at render time, as loading attachment is asynchronous we don't have the URL available at this point), but people also email around `spotify:` links and `itmss:` links to deep link into Spotify and/or the app-store.

I can understand the benefit of having a whitelist in many cases, but for us it just breaks too many things, and we'd like to go back to the blacklist approach.

I'm happy to add an option for this, but I think it would be better to go back to blacklisting anything that contains `script` verbatim.

ConradIrwin:


cure53:
Tests are green, code looks good, even the doc are updated, thanks for the super-thorough PR! :)

PR #147, opened 2016-04-13
make DATA_URI_TAGS configurable

jampy:
This commit allows to override or enhance the DATA_URI_TAGS list.

### DIFF
--- a/src/purify.js
+++ b/src/purify.js
@@ -250,7 +250,8 @@
     ]);
 
     /* Tags that are safe for data: URIs */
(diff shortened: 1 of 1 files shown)


cure53:
Looking good, but a test case would also be needed. Somewhere around [here](https://github.com/cure53/DOMPurify/blob/master/test/test-suite.js#L151) might be good :)

cure53:
Oh, and the README would have to be updated as well for completeness sake.

cure53:
Heya, just wondering - are you planning to add a test to this PR? If so, happy to merge.

Issue #129, opened 2015-12-08
Run node.js with jsdom in the browser grid


samertm:
Following up with https://github.com/cure53/DOMPurify/issues/61.

How much work would it take to run DOMPurify's test suite with node.js and jsdom? I'd be willing to help out.

cure53:
Good question :) What exactly do you have in mind?

samertm:
Ideally, one of the `customLaunchers` from test/karma.conf.js would be 'node_js_with_jsdom'. I have no idea how that works, and it's probably not worth doing unless jsdom passes the full test suite.

So I guess the best place to start is running the test suite with node.js. I some trouble adapting @Joris-van-der-Wel's script (https://joris-van-der-wel.github.io/DOMPurify-test-jsdom.js) to work with the new framework. I can continue working on that, but any pointers would be greatly appreciated!

Joris-van-der-Wel:
I have not checked in a while, however previously all tests were passing in jsdom, except for:

* the KEEP_CONTENT option, 
* A few `[OverrideBuiltins]` tests. These tests failed because jsdom does not implement a few properties that the tests check for. However those properties are not used by dompurify.

I do not think anyone has made a karma plugin for jsdom. I am not sure if it would make much sense to make one, you would add a lot of useless overhead.

Something I like to do is that I define two tasks using grunt. One task tests jsdom, the other task runs browserify and then karma for firefox, chrome, etc. (I am working on setting this up for jsdom itself too). A pattern I often use is this (example is using mocha):

```javascript
describe('my dom lib', function() {
	var document;
	var window;

	// `beforeEach` could also be used, this would give you a fresh document for each test.
	before(function() {
		// ( browserify automatically defines `global` to be `window` )
		document = global.document || require('jsdom').jsdom('<html/>');
		window = document.defaultView;
	});

	it('should do something fancy', function() {
		var myLib = new MyLib(window);
		if (myLib.foo(10) !== 20) {
			throw Error('uh o');
		}
	});
});
```

cure53:
Is there anything we can do at this point to help?

Joris-van-der-Wel:
Help with jsdom, or help with getting the test suite to work?

cure53:
With getting the test suite to work :)

samertm:
[...]

PR #149, opened 2016-04-23
Also run unit tests using jsdom + Node.js

Joris-van-der-Wel:
Also see #129

There is one failing test because of https://github.com/tmpvar/jsdom/issues/1460, which is why I have not yet added this new task to the CI script.

The new file `test/jsdom-node.js` is using ES6 syntax because jsdom requires that same syntax, so this test suite will fail early if someone tries to run it with a node.js version that is too old. I could easily change this if you all would like to enforce ES5 as a style requirement.

I am not that familiar with QUnit, so could someone double check that part? For example, is the `qunit` npm package a good runner for it? (I have noticed that it uses an older version of `qunitjs`, so we end up with two versions of `qunitjs`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "minify": "scripts/minify.sh",
     "amend-minified": "scripts/amend-minified.sh",
     "test": "npm run jshint && npm run-script travis-ci",
-    "travis-ci": "[ \"${TRAVIS_PULL_REQUEST}\" = \"false\" ] && ./node_modules/.bin/karma start test/karma.conf.js --log-level warn --reporters dots --single-run || false",
+    "travis-ci": "[ \"${TRAVIS_PULL_REQUEST}\" != \"false\" ] || ./node_modules/.bin/karma start test/karma.conf.js --log-level warn --reporters dots --single-run || false",
     "ci-test": "./node_modules/.bin/karma start test/karma.conf.js --single-run",
     "local-test": "npm run jshint;./node_modules/.bin/karma start test/karma.conf.js --browsers Firefox,Chrome --single-run",
     "jsdom-test": "node test/jsdom-node-runner"
--- a/package.json
+++ b/package.json
@@ -8,7 +8,8 @@
     "test": "npm run jshint && npm run-script travis-ci",
     "travis-ci": "[ \"${TRAVIS_PULL_REQUEST}\" = \"false\" ] && ./node_modules/.bin/karma start test/karma.conf.js --log-level warn --reporters dots --single-run || false",
     "ci-test": "./node_modules/.bin/karma start test/karma.conf.js --single-run",
-    "local-test": "npm run jshint;./node_modules/.bin/karma start test/karma.conf.js --browsers Firefox,Chrome --single-run"
+    "local-test": "npm run jshint;./node_modules/.bin/karma start test/karma.conf.js --browsers Firefox,Chrome --single-run",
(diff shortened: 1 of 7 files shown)


Joris-van-der-Wel:
Another thing to consider is that with jsdom we can initialize DOMPurify in two ways:

1\. Pass DOMPurify as a script to `jsdom.env()`, this will load DOMPurify running within jsdom's sandbox, just like in other browsers:

```javascript
jsdom.env({
    html: '',
    scripts: ['path/to/dompurify.js'],
    done(err, window) {
        window.DOMPurify.sanitize(...);
    }
})
```

2\. Initialize DOMPurify by passing jsdom's window object to the factory function. This is easier because you can do it synchronously:

```javascript
const window = jsdom.jsdom().defaultView;
const DOMPurify = require('DOMPurify')(window);
DOMPurify.sanitize(...);
```

I am using option 2 in this PR because this is more interesting to test, since option 1 is already tested in other browsers. We can test both methods, however I am not sure this is worth the extra cpu cycles.

Joris-van-der-Wel:
Seems like the `qunit` package indeed has issues. Some of the asyncTests are not running, I will look at it further later on.

Joris-van-der-Wel:
I've replace `qunit` (the runner) with `qunit-tap` so that we can use whatever version of `qunitjs` we'd like. The async tests are now running properly.

cure53:
Hi @Joris-van-der-Wel - wow, that's one major contribution, thank you :D Is the PR merge-ready? I reviews the changes, no objections from my side.

Thanks again!

Joris-van-der-Wel:
I corrected one more issue and it should now be merge ready. 

There are 4 failing tests, all with the same payload: `<div id="12"><body onscroll=alert(12)><br><br><br><br><br><br>...<br><br><br><br><input autofocus>//["'`-->]]>]</div>`, this is the jsdom bug i mentioned earlier.

cure53:
I am wondering if I can help somehow with the failing tests. We're planning to release 0.7.5 soon and it would be great to have this one on board as well :D

Joris-van-der-Wel:
I might have some time to look at it tomorrow (everyone gets the day off because its the kings birthday), however jsdom releases on weekends usually. I'm not sure yet how complicated the clean solution for jsdom is however, I'll have to dig into the HTML LS.

The issue (https://github.com/tmpvar/jsdom/issues/1460) is that if you set `document.body.onscroll = foo`, the onscroll event should also be set on the `window`, however in DOMPurify the documents have no window (because of createHTMLDocument). There are more spec violations going on in that piece of code though, so I want to pick those up too.

cure53:
Ooh I understand. Interesting! 

Then I would suggest we wait a bit with 0.7.5. Wouldn't make sense to go without that change. Is there anything we can do in our code base to make things better?

Joris-van-der-Wel:
Not really, this occurs during parsing of the html. So the Error is thrown while setting outerHTML. Anyway feel free to submit a PR to jsdom yourself if you'd like to work on it :p

Joris-van-der-Wel:
Should be fixed in jsdom 8.5.0. It would be cool if we could set up jsdom in DOMPurify's build script. I noticed that tests are currently not being run for Pull Requests, this is probably because of a SauceLabs permission issue? Jsdom would probably work for PRs, because those tests run completely on travis itself, without SauceLabs.

cure53:
@Joris-van-der-Wel Okay, what needs to be done for this?

PR #153, opened 2016-05-09
WIP: Fix build

fhemberger:
Ref: https://github.com/cure53/DOMPurify/commit/6e6157fc76842e68828850a483bac921b22f5508#commitcomment-17393154


fhemberger:
Okay, what I found out so far:
- Tests in isolation seem to run fine
- Running the tasks in the `local-tests` script work fine on their own: `npm run jshint; npm run jsdom-test; node_modules/.bin/karma start test/karma.conf.js --browsers Firefox,Chrome --single-run`
- But running the same local tests with `npm run local-tests` fails with `ELIFETIME`.

Tested with npm 3.8.6, node v6.1.0, other versions seem to be affected in the same way.

I also found something odd in package.json, which I cannot explain: Why are we using https://github.com/shirish87/karma-browserstack-launcher/tree/global_poll_0.1.6
instead of the original project, and why this particular branch? Maybe there are some upstream fixes as well.

cure53:
> I also found something odd in package.json, which I cannot explain: Why are we using https://github.com/shirish87/karma-browserstack-launcher/tree/global_poll_0.1.6

We had some problems with the Karma runner and Browserstack and the folks from Browserstack recommended to use this particular version.

cure53:
Hi @Joris-van-der-Wel! We seem to have run into some problems with the tests that we cannot seem to be able to resolve. Would you mind having a look please?

fhemberger:
Aaaahhh … I intentionally labeled the PR as "WIP" so you don't merge it. ;)
I'm still trying to find out what's breaking the tests. Maybe we should also check if the Browserstack issues have been fixed in the meantime.

cure53:
@fhemberger The builds are broken anyway - we cannot break them even more :)

Joris-van-der-Wel:
@cure53 do you have a link to a build that failed?

cure53:
https://travis-ci.org/cure53/DOMPurify :)

Joris-van-der-Wel:
This is caused by a regression in jsdom: https://github.com/tmpvar/jsdom/issues/1477. This PR updated jsdom from version 8 to 9.

fhemberger:
@Joris-van-der-Wel Build was already broken in the last commit, without touching jsdom. The jsdom part works, it seems to be something with karma/browserstack.

PR #154, opened 2016-05-12
WIP: Test for failing builds

fhemberger:


fhemberger:
'push' tests don't run the latest changes (so it fails), but the general PR seems to work.

cure53:
Ready for a merge?

fhemberger:
As you said the last time: The build is already broken ;)

This should hopefully improve a few things, but I think in general the whole Karma/BrowerStack tests are borked. I now re-ran them a couple of times on Travis: Sometimes they pass, sometimes they don't (throwing one or more of: "browserstack binary not installed", "browser.on is not a function", "test caused full page reload", "disconnected", etc).

cure53:
So, what do we do about those?

fhemberger:
Honestly, I don't know right now. As I'm also not the person who set up this testing infrastructure in the first place.

cure53:
@Joris-van-der-Wel and @tdeekens do you have any ideas what could be the breakage here?

fhemberger:
Great, looks we hit BrowserStack's rate limit this time. m(
Is there a more stable service we can use for browser testing?

cure53:
We already got a bigger request limit than usual from them. Unlikely they will give us more (for free).

fhemberger:
As a couple of well-known companies are using DOMPurify, how about asking them for sponsoring our test setup?

Joris-van-der-Wel:
@cure53 Well, the jsdom breakage is fixed for now because the version was reverted to 7. I see that in the travis job log that it goes through the jsdom tests without error now. (note that the build script will not start the karma test if jsdom fails).

Are you sure the error from browserstack is not just it complaining about retrying too fast? Is the DISCONNECTED error not the real error, or is that also caused by the rate limit?

tdeekens:
I've had the same issues in others projects and they don't even always relate to BrowserStack and e.g. SourceLabs won't make our life easier.

> "browserstack binary not installed", "browser.on is not a function"

Are definitely bound to it and I will have a look as soon as I can.

> "test caused full page reload", "disconnected"

Are more rooted in the nature and structure of the test suite imho.

PR #163, opened 2016-05-26
fix a bug

kamael:
if `dirty` start with a tag(eg. `<style>`) which will be parsed into head by DOMParser,  the tag wiil disappear in result.

```
var dirty = '<style f="x"></style>aaa<style f="y"></style>';
const clean = DOMPurify.sanitize(dirty));
/*
 except       <style f="x"></style>aaa<style f="y"></style>
 but shows    aaa<style f="y"></style>
*/
```

### DIFF
--- a/src/purify.js
+++ b/src/purify.js
@@ -377,6 +377,8 @@
     var _initDocument = function(dirty) {
         /* Create a HTML document using DOMParser */
         var doc, body;
+
+        dirty = '<div></div>' + dirty;
         try {
             doc = new DOMParser().parseFromString(dirty, 'text/html');
         } catch (e) {}
@@ -390,6 +392,8 @@
             body.parentNode.removeChild(body.parentNode.firstElementChild);
             body.outerHTML = dirty;
         }
+
+        doc.body.removeChild(doc.body.firstElementChild);
 
         /* Work on whole document or just its body */
         if (typeof doc.getElementsByTagName === 'function') {


cure53:
Heya, thanks for the bug fix. However, it's not working for us. Injecting a random DIV into the content before building it interferes with DOMPurify in too many ways.

I recommend to simply use this setting:

```javascript
var clean = DOMPurify.sanitize('<style>hello</style>', {WHOLE_DOCUMENT: true});
console.log(clean)
```

Then you get everything as you probably need it - no auto-hoisting of style elements into the page header (and no disappearing styles if no header is returned).

cure53:
I reviewed the possibility of injecting and then removing some content prior and after document creation - but I think it's not a good way to solve this.

I guess, the path using `WHOLE_DOCUMENT: true` is the ideal one. Or is there anything that is a blocker for you?

kamael:
WHOLE_DOCUMENT will return all tags, contains html, head, body 
but the dirty is only inserted into some tag in body , like a div.

cure53:
Yep, I know. But working around the browser's internal parsing and defining our own special treatment for style elements would do more harm than good.

If, in your case, the behavior as it is causes problems, you can easily just add the initial element yourself and strip it out after you get the sanitized HTML back. Basically use the same trick that is in your pull request "locally", by prepending `<div />` before sanitizing and removing it right afterwards.

Yet, changing the core of the library and affecting everyones results by changing the parser rules is not gonna happen :D

PR #166, opened 2016-06-06
Attr security flaw

neilj:
2 minor fixes and one major security patch:

The logic for whether to forbid an attribute was broken: if
ALLOW_UNKNOWN_PROTOCOLS was true, then it would allow any attribute with
a value that didn't match the IS_SCRIPT_OR_DATA regexp.

This evades the attribute name whitelist, and is a security flaw: for
example you could set any "onFoo" JavaScript attribute.

The logic here got too convoluted to be easy to read, which is
presumably how this bug crept in. I have rewritten it into an expanded
set of if/else statements which I believe are easier to follow.

### DIFF
--- a/src/purify.js
+++ b/src/purify.js
@@ -835,7 +835,7 @@
      * @return void
      */
     DOMPurify.removeAllHooks = function() {
-        hooks = [];
+        hooks = {};
     };
 
     return DOMPurify;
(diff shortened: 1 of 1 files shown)


cure53:
Wow, awesome find! Thanks a lot :D Merging now. We should release 0.8.1 soon after.

PR #167, opened 2016-06-07
Fix DOMPurify.removed for SAFE_FOR_JQUERY mode

dnkolegov:
Additional check in `SAFE_FOR_JQUERY` mode was deleted in `0.8.1` version. As result `DOMPurify.removed` contained elements even if input data is valid. I have restored one and added some new tests.


cure53:
Yeah, I had to remove the check yesterday, because it broke too many tests and the security fix had priority. I will review the PR later today.

cure53:
Alright, looking good :) Thanks - and sorry for the hiccups. We had to address the security issue yesterday with highest prio - and this one fell under the table.

dnkolegov:
Cool. Thanks. Our internal tests  for `DOMPurify` in waf mode are successful.

cure53:
:+1:

PR #173, opened 2016-07-28
Reduce package footprint

fhemberger:
Removes tests, demos, build scripts, etc. from package manager distribution
to keep the overall file size down and improve installation speed.

- package.json: Only include files in 'src' and 'dist' in npm
- bower.json: Ignore all files except for 'src' and 'dist' for bower

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -24,7 +24,9 @@
   ],
   "ignore": [
     "**/.*",
+    "demos",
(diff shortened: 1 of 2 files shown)


tdeekens:
Looks good.

cure53:
Merged, thanks a bunch!

PR #191, opened 2016-12-01
Allow users to pass DOM nodes for sanitization

### LINKED ISSUE #190: Can't sanitize some DOM snippets, e.g. <td>text</td>
input: `<td>text</td>`
expected: `<td>text</td>`
actual: `text`

edg2s:
Fixes #190.


cure53:
Thanks! 

The PR looks good - we'll clean it up a bit after merging though (missing test cases, missing path for sanitizing a Node on older MSIE etc.).

cure53:
The code breaks MSIE, trying to debug...

edg2s:
Oh dear, which version?

cure53:
All of them :) I am on it...

cure53:
Okay, I made some changes that should fix it for MSIE11 and all older versions. Now adding some tests.

edg2s:
Out of interest why do mark IE9 as incompatible?

cure53:
Because MSIE9 has severe bugs in the markup processor that practically forbid using the DOM to sanitize. DOMPurify still delivers for IE5-IE9 by making use of `toStaticHTML`.

cure53:
@edg2s So I created one first test case and the results look good:

```javascript
  // Test to make sure that the node scanning feature delivers acurate results on all browsers
  QUnit.test( 'DOMPurify should deliver acurate results when sanitizing nodes', function (assert) {
      var clean = DOMPurify.sanitize(document.createElement('td'));
      assert.equal(clean, "<td></td>");
  } );
```

Can you come up with other examples an use-cases? i.e. from mentioned Visual Editor?

edg2s:
`<td>` is the go-to case for re-parenting (and other table nodes, e.g. `<tr>`). You could also test other node types, e.g. text nodes/comment nodes. There's not really any sanitsation to do there but the function should allow them.

I put in a special case for a `<body>` tag being passed in, but it may not work with an `<html>` tag

I also note that it might not work as expected if the full document return options are used, so those may be worth testing.

edg2s:
On the IE9 issue, I'm aware that DOMParser doesn't support text/html, but we have some fallback methods: https://github.com/wikimedia/VisualEditor/blob/master/src/ve.utils.js#L812

Are there other serious parsing issues once you have the DOM nodes?

If not it may be worth separating out "isSupported" into "canSanitize" and "canParseStrings", so if an IE9 user passes in DOMElements, it doesn't give up.

cure53:
> On the IE9 issue, I'm aware that DOMParser doesn't support text/html

It's not about that. There is other serious issues. Check this example in IE9: `<%/z=%&gt&lt;p/&#111;nresize&#x3d;alert(1)//>`

Right now, DOMPurify allows passing in DOM nodes. But it returns a string. This is not ideal yet of course but likely can be fixed.

PR #197, opened 2017-01-16
Fix removing data:image within src having whitespace

GreLI:
I have an email with HTML string like:
```html
            <img src="
data:image/png;base64,iVBORw0K...
```
and it got `src` removed due to whitespace symbols `\r\n` before `data:image/png`. This happens because of comparing index with 0 in [purify.js at line 642](https://github.com/cure53/DOMPurify/blob/1b72c3e/src/purify.js#L642):
```js
value.indexOf('data:') === 0 &&
```
I've added `value.trim()` to make this check pass. It's an ECMAScript 5 method, which I think is ok, since you're already using `.foreach()`.

Some tests are failing on my local run, but not the added one.


cure53:
Heya, thanks a lot! The change looks good! I'll merge and run our tests to see if all is fine. Should be though.

PR #209, opened 2017-04-23
Cleanup

neilj:
This mainly tidies code. Other than that, there are two real changes:

1. Add `template` to the list of `FORBID_CONTENTS` tags. If you strip the template, you should always be stripping everything inside it too. It makes no sense to keep stuff, since these would not have been rendered in the full document.

2. Remove the fallback for IE8/9 and throw an error instead if DOMPurify is not supported. The current behaviour is actively dangerous, because it silently either doesn't do anything, or behaves differently to what is expected (for example none of the config options are respected). Throwing an error forces developers who care about supporting older browsers (essentially IE<9) to explicitly check if DOMPurify is supported and actually think about an alternative codepath. For developers who don't care, breaking is fine and probably still better than a security hole.


tdeekens:
Removing the fallback is a breaking change, correct?
On another side note: do we have any opinion about our initialsation logic with the factory function. The pattern of the DOMPurify calling its UMD factory function makes code modularisation and refactoring very hard. Removing it would however also be a breaking change most likely.

cure53:
I think this is fine aside from https://github.com/cure53/DOMPurify/pull/209/commits/cff54ec8987e1217e26ec3c6b32785b91f3b5618.

We should think about doing this for 1.0 at some point but not now.

cure53:
Also we have to bring DOMParser back because of a parser bug in Firefox (that FF devs have been ignoring for a couple of months now).

neilj:
OK, I've removed that commit from this branch. I think it's the right thing to do, but sure leave it for 1.0. It is kind of a breaking change, but to be honest it's already broken if you use those browsers. Better you make sure the developer is aware of it in my opinion!

Regarding DOMParser, yeh I was going through the history to see why it was introduced in the first place and [found this]( https://github.com/cure53/DOMPurify/releases/tag/0.6.7) so was going to email you about that. Do you have some good test cases for this? All the tests are currently passing so I presume there are no tests that cover this specific scenario at the moment.

cure53:
Thanks :) I am currently implementing a tweak to get around the DOMParser issue but I really don't like it so far, basically it involved Safari detection via DOM property.

I hope they fix the issue soon and we can get rid of that part of the code. Alternative ideas are welcome too.

cure53:
Okay, my changes are in. Should be okay.

cure53:
I think we should also issue a new release just in case. Thoughts?

neilj:
This patch is broken in iOS – Safari has the same bug, but there is no `window.safari` property. Do not release this!

cure53:
Oh no - that's bad! Is there any unique DOM objects we can use for browser detection on iOS Safari? I tested OSX Safari and it looked reliable.

neilj:
I'd much rather see a solution that did not reintroduce `DOMParser`. Do you have a test case for the Firefox bug? Maybe we can try some other ways around it.

cure53:
The problem with Firefox is, that we cannot trust innerHTML anymore when dealing with SVG. A ticket was filed ages ago but I doubt it will ever be addressed. Here is more info:

https://github.com/cure53/DOMPurify/releases/tag/0.6.7

DOMParser is in my opinion the right way to go. But we can of course also try to find a way around the FF issue and be fine with that.

neilj:
Hmm, I've been unable to come up with a workaround. I think I am reluctantly agreeing that we have to use DOMParser for Firefox. I think the best solution is to feature detect the defect in Firefox's innerHTML – anything else is going to be too brittle and liable to cause security issues. I've created a pull request for this.

PR #210, opened 2017-04-24
Feature detect when to use DOMParser

neilj:
It's safer to feature detect where possible, which we can do for the Firefox bug but not for the Safari bug. So let's default to not using `DOMParser`, unless we detect we need it.

I'm still not 100% happy – the reason the `DOMParser` code was [added in the first place](https://github.com/cure53/DOMPurify/releases/tag/0.6.7) was because `innerHTML` in FF uses a different parser when compared to `document.write` (it seems like `innerHTML` may always use the quirks-mode parser?), and so this could introduce a bug if you sanitised the DOM tree generated from this, then used the HTML representation with `document.write`. But by switching to the other parser, it strikes me there may be a vector out there now which parses fine with `DOMParser` but will cause a security hole when passed to `innerHTML`! I don't have such a vector, but the fact there's a parser difference means it might exist. I don't know there's anything we can do about it though, but I just wanted to mention it.

### DIFF
--- a/src/purify.js
+++ b/src/purify.js
@@ -46,8 +46,7 @@
     var Text = window.Text;
     var Comment = window.Comment;
     var DOMParser = window.DOMParser;
-    var safari = (window.safari && 
-        typeof window.safari.pushNotification === 'object') || false;
+    var useDOMParser = false; // See comment below
 
     // As per issue #47, the web-components registry is inherited by a
     // new document created via createHTMLDocument. As per the spec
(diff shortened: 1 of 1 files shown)


cure53:
Checking for Firefox instead of Safari looks good! However, should we not fingerprint for `window.dump` or `window.uneval` instead? 

If they are native functions, we can opt into DOMParser, if not, we can go the other path? Might be faster than actually creating the whole document.

neilj:
The trouble with this is that other scripts could add variables with those names into a page in Safari. Or  could write them to null in Firefox. Detecting like that is not safe.

cure53:
How about this then?

`document.implementation.createHTMLDocument('').onafterscriptexecute`

Should be tamper-safe and clearly identify Firefox. https://developer.mozilla.org/en-US/docs/Web/API/Document/onafterscriptexecute

neilj:
That would probably work, but I don't get why you would do that instead of actually detecting the bug we're working around. If Firefox removed `onafterscriptexecute` in the future, the vulnerability would be reintroduced. If Firefox fix the bug in the future, the code is still fine!

cure53:
Yeah, you are right, this might happen. I have an eye on these changes but might miss it and then things will break. I think the PR is good!

cure53:
I think we are ready for a new release, any objections?

PR #206, opened 2017-03-18
Refactor library to support ES201* JavaScript

tdeekens:
> This adds rollup as a build tool and relates to issue #202.

### Background & Context

Currently DOMPurity does not have a built tool in place. This is nice and often simplifies things. However, adding webpack might give the project some benefits and this issue is about to align on the potential gain of it.

This PR might change quite a couple of things and can/should be seen as a start of a discussion on how far to go with things.

1. Change in API with more granular exports (not "just" the object) e.g. sanitize and
   - to allow `import sanitize, { addHook } from 'purify'` 

### Included in this

1. Adding webpack with configuration for umd build
2. Removing jshint replacing it with xo
   - to allow ES201x syntax to be linted with a default set of sane eslint rules
3. Refactors some code to illustrate potential in changes
   - Some files where put into separate modules
   - Other changes fix linting errors while others add exceptions
4. Rewires jsdom to test built version of purify

### Open complications

1. Mode detection (running in browser or node)
   - exports are static and early return won't do (?) which breaks script parsing in node

I tried to split up changes into sensible commits to be "easier" to digest. I hope it worked. 

closes #202

PR #211, opened 2017-04-27
Export in global scope

mkazlauskas:
> This pull request is for Electron, where it won't export to global scope when imported as <script>

### Background & Context

Also removes deprecated version field in bower.json


cure53:
I think the version numbers are highly outdated :)

@fhemberger What do you think about the other change? I lack expertise here to judge it properly.

src/purify.js
```diff
@@ -7,9 +7,8 @@
         define(function(){ return factory(root); });
     } else if (typeof module !== 'undefined') {
         module.exports = factory(root);
-    } else {
-        root.DOMPurify = factory(root);
     }
+    root.DOMPurify = factory(root);
```
fhemberger: Tests are failing:

```
/home/travis/build/cure53/DOMPurify/src/purify.js:11
    root.DOMPurify = factory(root);
                   ^
TypeError: Cannot set property 'DOMPurify' of null
    at /home/travis/build/cure53/DOMPurify/src/purify.js:11:20
    at Object.<anonymous> (/home/travis/build/cure53/DOMPurify/src/purify.js:12:2)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Module.require (module.js:367:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/home/travis/build/cure53/DOMPurify/test/jsdom-node.js:7:17)
    at Module._compile (module.js:413:34)
```

fhemberger: I don't know what the issue is with Electron. Can you please elaborate a bit more?

cure53:
Any info or news here? @mkazlauskas

mkazlauskas:
I'm not sure how your tests fail. What my change intended to do is export the library to global scope (window) in Electron environment in the same way as in regular browser environment. Since Electron has both `require` and `window` it kinda makes sense to export to both. Also, in my specific use case I must use it from global environment and can't use require (doing HTML Imports to load it).

cure53:
cc @fhemberger @neilj

cure53:
Given the massive changes in #206, does this PR still make sense?

src/purify.js
tdeekens: We should not interact with the `factory` of the surrounding UMD build in the first place. This has changed after merging #206 where we export a function `createDOMPurify` function you can use to wrap the library onto another global. You can see an example in our test setup for jsdom [here](https://github.com/cure53/DOMPurify/blob/master/test/jsdom-node.js#L6).

cure53:
Closing this for now, feel free to re-open if this still makes sense.

PR #212, opened 2017-05-03
Fix broken test for useDOMParser

neilj:
The logic was inverted, so this was broken in *both* the Safari and Firefox
edge cases.

This replaces it with a simpler test, which I have confirmed is right in both Safari and Firefox. One question though: why was this not picked up by the automated tests? The test suite should have failed in Safari 10.1 with 0.8.7, and I thought you had added this to the browser matrix?


cure53:
> and I thought you had added this to the browser matrix?

Not available yet on BrowserStack.

Issue #199, opened 2017-02-21
Extra <body> node and ignored <script> with hook uponSanitizeElement


jfparadis:
Two issues with hook uponSanitizeElement:
- Calling sanitize causes one extra callback, aways with extra <body> node. 
Example: `<div><script>alert(0)</script></div>`
Hook called 4 times: `BODY, DIV, SCRIPT, text`

- When sanitizing string containing only <script> tag, there is no callback for script.
Example: `<script>alert(0)</script>`
Hook called 1 time: `BODY`

Please see this: [JSFiddle](https://jsfiddle.net/2L3rjng4).

cure53:
Ah, that's the old problem with the browser moving SCRIPT, STYLE and others into the HEAD element when not prepended by any other element. This is causing trouble in many situations. Technically not our fault but I feel more and more we should fix this.

How do you feel about a `FORCE_BODY` configuration flag that simply makes sure that nothing goes into the HEAD? We could default it to `false` right now for 0.8.x and set it to `true` starting with 0.9.x.

jfparadis:
Very interesting. 

Question: if <script> gets moved to <head>, why uponSanitizeElement is not called with <head>, then with the <script>?

cure53:
We use the `doc.body` as the platform to check and sanitize the markup :) That's why it simply disappears.

cure53:
Would the proposed fix work for you?

jfparadis:
Yes, `FORCE_BODY` would work. There is another approach that mimic the browser and could be simpler. Here is the idea:

If I create a simple `test.html` with the following content:

```
<script src="hello.js"></script>
```

And if I load it in the browser and look at the DOM, I obtain:

```
<html>
    <head>
        <script src="hello.js"></script>
    </head>
    <body>
    </body>
</html>
````

Because the callback starts at `doc.body`, we get the current behavior (the head is ignored). 

Now, if I create a simple `test2.html` with the following content:

```
<html>
    <head>
        <script src="hello.js"></script>
    </head>
    <body>
    </body>
</html>
````

I get the same behavior, the `<head>` is also ignored ( you can see the result on my updated [JSFiddle](https://jsfiddle.net/n0w7961L/).

An alternative approach would be to call `doc.head` followed by `doc.body`. The advantage would be to always process head markup from body markup, and to mimic more closely the browser behavior.

cure53:
[...]

PR #213, opened 2017-05-14
Option to set config once

jfparadis:
This pull request adds documentation and tests to complement:
https://github.com/cure53/DOMPurify/pull/199

Since the original request, I evaluated a different API based on a push/pop principle to be more symmetric with addHook/removeHook, but merging configurations required more code, and my take is that it made the system too complex and not required for current use case. 

The idea was to keep an array of cumulative configurations, and re-pase on push/pop. The extra code was required for deep cloning and deep merging the config objects. I can do the work if you prefer to take that route.


cure53:
That looks good from my side! Thanks :)

@neilj @fhemberger any objections from your side?

jfparadis:
Thank @cure53. Are there change required?

cure53:
I think we are good here, thx :)

PR #218, opened 2017-05-25
Fix and refactor karma config and setup with rollup

tdeekens:
> This pull request fixes karma setup with rollup.

### Background & Context

1. Moves the specific rollup setup for karma to the karma.config.js
2. Fixes using relative imports for the test setup

Just another try. It's really a bit of poking. I will have to leave the keyboard after this until tonight. I hope that's okay but I'll get to it again then.


tdeekens:
Maybe a take away from this is that we run at least one browser on BS when testing PRs?

cure53:
Trying :D

cure53:
It's working :D

tdeekens:
...and I dont want to know why.

PR #221, opened 2017-05-31
Fix purify `module.export` over `export default`

### LINKED ISSUE #220: dist/purify.js is broken
The current `dist/purify.js` file **does not work** in a browser. It tries to assign to `module.exports` and `module` is undefined. Was this really not tested before the PR was merged?

tdeekens:
> This pull request fixes using an ESM default export over a CJS counter part.

⚠️ Please read and verify the below mentioned before merging ⚠️ 

### Background & Context

The module should, left alone for the sake of modernity, export itself through ESM. We currently only have one export `default` which before was the return of the `UMD` closure and then became the `module.exports`. The `module.exports` is currently no working in browsers and passes the tests as `rollup` is instrumented to be able to parse CJS modules (to setup the tests).

The CJS export was chosen for a varity of reasons

1. When importing/executing DOMPurify we want it to expose itself as a global (through the UMD bundle)
2. When importing DOMPurify in a node.js context we don't want to have to `require('DOMPurify').default`. 
3. When importing DOMPurify through a bundler like webpack/rollup/browserity we still want the library the work

Generally to support all those the community used to use UMD but slowly moves towards having targeted bundles offered per environment.

Just changing the `module.exports` to an `export default` seems to work. It still has to be verified that `DOMPurify` exposes itself on the `window`.

### Tasks

- Replace CJS `module.exports` for ESM `export default`

### Dependencies

Closes #220


tdeekens:
Please test on the branch next time. Now we have to swing up another PR etc :)

cure53:
So, what now? Can I help anything?

tdeekens:
...next one. This works for me with the `demos/basic-example`. For you too?

tdeekens:
https://github.com/cure53/DOMPurify/pull/222

PR #227, opened 2017-06-28
Updated example of JSDOM usage on Node.js

ydaniv:


cure53:
Looks good from my side, thanks! What do you think @tdeekens ?

Issue #230, opened 2017-07-06
New profiles remove valid attributes and tags


ydaniv:
Now with new profiles, attributes that are valid in more that one profile (e.g. `width` and `height`) and would have been duplicates, are now removed when selecting a single profile.

E.g.:
This:
```
<svg viewBox="0 0 120 120">
    <rect y="10" x="10" width="50" height="100"/>
</svg>
```
Becomes:
```
<svg viewBox="0 0 120 120">
    <rect y="10" x="10"/>
</svg>
```
Where expected is:
```
<svg viewBox="0 0 120 120">
    <rect y="10" x="10" width="50" height="100"/>
</svg>
```

So when using profiles, need to make sure that valid attributes from excluded profiles are whitelisted for included profiles where they should be.

Thanks!
~Y

cure53:
There is two ways we can fix this:

 * Create a file with shared attributes
 * Extend the existing collections and have duplicate attributes

I did the last one, PR from you is welcome :)

ydaniv:
I'm on it (: 👍

cure53:
Awesome, thx! :)

ydaniv:
BTW, I'm using [this page](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute) as references, and I see some missing, I guess it's mostly on purpose or stuff related to SVG2, so currently not adding new ones.

Let me know if you'd like otherwise.

cure53:
Yeah, some of fancy SVG2-ones we don't have yet.

ydaniv:
OK, so not adding anything new.

Since attribute lookup is already implemented as a Set it's trivial to just add duplicates in `attrs.js`.
So I think I've covered all of them.

cure53:
Yeah, we can check after this PR came through what else is harmless and worth adding.

cure53:
Okay, local tests look fine, let's see what Browserstack says :)

ydaniv:
Woops, problemo. We need to do the same for tags. E.g. it's removing `style` tags from SVG in non HTML profiles.

Shall I open a new issue/PR?

cure53:
Yes, plz!

PR #232, opened 2017-07-09
See #230

ydaniv:


cure53:
Looking good but no tests. Cannot accept the PR without :)

ydaniv:
Hmmm... will do. But notice it's failing on `lint`. It's weird, since I didn't pass it at first and couldn't commit, on stuff in `purify.js`, which I didn't touch, and then ran `format`, passed `lint` and pushed. And now, I don't the CI is breaking on the same rules in lint.

ydaniv:
So, added 1 test, which I think covers it 😕 
But still failing on that discrepancy in lint config.

tdeekens:
Sounds like conflicting prettier and eslint versions or rules due to unpinned dependencies maybe. Sorry for the inconvienence. Just code on and I'll check things as soon as I find time.

tdeekens:
Just reminded myself that I added yarn to this project. The lockfile should pin everything. Did you install dependencies via yarn? We might have to add an npm-shrinkwrap or lockfile while deciding on what npm version we want to support.

ydaniv:
npm. a note on the README would be nice 😃

tdeekens:
Sure. @ydaniv what do you think of #235. Would that have clarified things for you?

cure53:
Also, https://github.com/cure53/DOMPurify/commit/2dd0206cb74a70dc567d251ff1143154ea85f5e8 :)

cure53:
Something seems wonky @tdeekens 

https://travis-ci.org/cure53/DOMPurify/jobs/253332077

PR #238, opened 2017-07-25
Added missing xmlns attribute in SVG

ydaniv:
This was missing in #232


ydaniv:
Basically means DOMPurify cleans `xmlns` attributes on `<svg>` when using `USE_PROFILES: {svg: true}`

PR #241, opened 2017-08-16
Fix bundling and testing

tdeekens:
> This pull request should fix the tests and bundling.

### Background & Context

Dove into rollup configs and setup for some while. This seems the easiest setup to get it all running again. There is a bazillion options but this seems like a good 80/20 for us. I got lost in optimizations around external-helpers for rollup and not using babelrcs of child projects but I think it all does not matter here.

It works local for jsdom and karma.


cure53:
Tests are green, let's see :) Thanks!

tdeekens:
Arr. The error isnt there locally. I am confused. Should be gone by only having bundle size on prod building.

PR #245, opened 2017-08-20
test: on firefox for all node versions

tdeekens:
### Background & Context

Will run FF tests on all node versions
Will run BS tests on master

### DIFF
--- a/test/karma.conf.js
+++ b/test/karma.conf.js
@@ -1,29 +1,9 @@
-const argv = require('minimist')(process.argv.slice(2));
-const isArray = require('lodash.isarray');
 const commonjs = require('rollup-plugin-commonjs');
 const includePaths = require('rollup-plugin-includepaths');
 const rollupConfig = require('../rollup.config.js');
 const customLaunchers = require('./karma.custom-launchers.config.js')
   .customLaunchers;
 const browsers = require('./karma.custom-launchers.config.js').browsers;
-
-/**
- * Environment variables are passed into the script and the depth of testing
- * is affected accordginly.
- *
- * - Whenever on a PR we only want to probe test with Firefox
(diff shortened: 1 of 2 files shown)


test/karma.conf.js
```diff
@@ -1,30 +1,10 @@
...
- * Environment variables are passed into the script and the depth of testing
- * is affected accordginly.
- *
- * - Whenever on a PR we only want to probe test with Firefox
- * - Whenever we are on the most recent node version on Travis we test via BrowserStack
- * - If none of the prior mentioned holds we assume to be running local and respect the passed
- *   in borwsers argv
- * - If we did not receive any argv browsers we succeed (other node versions on Travis)
- */
-const shouldProbeOnly = argv.shouldProbeOnly !== '';
```
tdeekens: This is just extracted over.

test/karma.custom-launchers.config.js
```diff
@@ -140,8 +16,27 @@ const customLaunchers = {
...
+ *
+ * - Whenever on a PR we only want to probe test with Firefox
+ * - Whenever we are on the most recent node version on Travis we test via BrowserStack
+ * - If none of the prior mentioned holds we assume to be running local and respect the passed
+ *   in borwsers argv
+ */
+const shouldProbeOnly = argv.shouldProbeOnly !== '';
+const shouldTestOnBrowserStack = argv.shouldTestOnBrowserStack !== '';
+const defaultBrowsers = ['Firefox'];
+const argvBrowsers = isArray(argv.browsers)
```
tdeekens: Defaulting to FF
Respecting whatever we get passed in via argv

If we want BS and we probe we run FF otherwise any browser
If we don't we just pick from argv with the same defaulting to FF

tdeekens:
Should be fine. Played with env var combinations locally.

Issue #249, opened 2017-08-21
Usage of DOMPurify requires "unsafe-eval" CSP since 1.0.0


frigus02:
As described in the [CSP spec](https://w3c.github.io/webappsec-csp/#directive-script-src), using `Function(...)` with a string requires the "unsafe-eval" CSP rule.

This is now the case with the implementation of `getGlobal`:

https://github.com/cure53/DOMPurify/blob/6185fe544a9fdd21453fbcddea608dcd35ace040/src/purify.js#L5-L8

I'm not sure why this pattern is used here. I guess you want to be sure to get the global object.

Do you think you can release a CSP compliant version of DOMPurify, which just returns `window`?

cure53:
Good point, can we work around that @tdeekens?

tdeekens:
As far as I remember it was needed to either get the `node.global` or the `window`. Suggestions welcome.

tdeekens:
Not sure if moving to https://github.com/purposeindustries/window-or-global/blob/master/lib/index.js would solve the issue while still be working.

cure53:
It would for sure not cause CSP warnings :D

tdeekens:
The problem with the library is that `this` will be rewritten to undefined using rollup (https://github.com/rollup/rollup/wiki/Troubleshooting#this-is-undefined). I am not really sure what the "semantics" behind the idea for "root" at say https://github.com/cure53/DOMPurify/blob/a992d3a75031cb8bb032e5ea8399ba972bdf9a65/src/purify.js#L1 which is where the `factory` was used and interacted with which we can't do now as we bundle the library up.

tdeekens:
Check https://github.com/cure53/DOMPurify/pull/250

tdeekens:
Build is green. I guess it's time for 1.0.2 after all.

frigus02:
Thanks again for the quick solution :-).

tdeekens:
You're welcome. I hope the solution covers all our use cases of consumers.

PR #250, opened 2017-08-21
Refactor `getGlobal` to not cause CSP warnings

tdeekens:
### Background & Context

This relates to #249.

Please critically reflect the old UMD interaction with the `factory` as in https://github.com/cure53/DOMPurify/blob/a992d3a75031cb8bb032e5ea8399ba972bdf9a65/src/purify.js#L1-L16 was there before I worked on the library.

Locally JSDOM and Karma against Chrome works fine.


frigus02:
Thanks for the nearly immediate response. You are amazingly fast.

Maybe just don't see it. But did you actually include and use `root.js` now?

tdeekens:
No, we can't use `window-or-root` for reasons stated in #249.

1. It falls back to `this` on a top level which rollup will replace to `undefined` due to https://github.com/rollup/rollup/wiki/Troubleshooting#this-is-undefined
2. It's not really working as stated in the docs and does not have an ESM build

The "real" change is https://github.com/cure53/DOMPurify/pull/250/files#diff-39d3e4cf739c51697e855107d73a23f5R5. However, I am not 100% sure if that covers all semantics/needs towards a `getRoot` for the library and an older maintainer should answer.

tdeekens:
A now I see your point. Sorry `root.js` was a left over from experimenting. Now removed.

cure53:
Aaaand merging!

PR #252, opened 2017-09-04
Feat/add new protocols

dhoko:
> This pull request adds new schemes

- callto
- cid
- xmpp

Tests :ok:


cure53:
Heya, can we remove blob? This would cause XSS and is considered unsafe.

tdeekens:
@dhoko let me know if there was any unclarity regarding build setup, test tooling or the readme. I'd love to improve it and ease contributions!

dhoko:
@cure53 np, do you have any details about the blob ?
@tdeekens  voilà ;) (_I tried to install_ `hustky` and `precommit`)

screenshot from 2017-09-05 10-00-48

cure53:
@dhoko Blob URIs run on the same origin as the creating website and, on Chrome, ignore CSP if present. Hence, they are almost equally dangerous as Data URIs - but require some framework magic to be in place to be exploitable in the first place (i.e. Angular 1.x).

Either way, too risky for the DOMPurify default.

tdeekens:
You mean husky? We use "pre-commit" might be a case for moving over...

dhoko:
@tdeekens yep, idk it didn't work with both of them :/

cure53:
With blob being removed, is this ready for a merge now? @tdeekens

tdeekens:
Yes, I might move us to husky but thats a separate thing.

cure53:
Thanks all :)

PR #253, opened 2017-09-05
Feat/extend protocols

dhoko:
> This pull request will allow the user to extend the protocol regexp in order to add a specific filter instead of ALLOW_UNKNOWN_PROTOCOLS (_everything_). 

Ex: I want to only allow `ftp`.

Tests are ok.


README.md
```diff
@@ -139,9 +139,14 @@ var clean = DOMPurify.sanitize(dirty, {ADD_ATTR: ['my-attr']});
 var clean = DOMPurify.sanitize(dirty, {ALLOW_DATA_ATTR: false});
 
 // allow external protocol handlers in URL attributes (default is false)
-// by default only http, https, ftp, ftps, tel and mailto are allowed.
+// by default only http, https, ftp, ftps, tel, mailto callto, cid and xmpp are allowed.
```
plobbes: nit: add a comma (,) after mailto

cure53:
I love it, thanks :D But since we introduce a new flag, we also need a test case here:

https://github.com/cure53/DOMPurify/blob/master/test/test-suite.js#L458

cure53:
Further we need to mention the new flag in README.md.

cure53:
Thanks a ton :)

PR #267, opened 2018-01-24
Update code example to show usage with older JSDOM API

Vector919:
Demonstrating DOMPurify usage with the older JSDOM API

> This pull request updates the readme to show an example of DOMPurify working with the older JSDOM API for server side rendering

### Background & Context

Should clarify usage for anyone still using the old JSDOM api.


cure53:
Love it, thanks!

Issue #269, opened 2018-03-19
svgFilters in attrs.js missing 4 filters


ydaniv:
Hey,

Question, I see 4 filters missing form `svgFilters`:

- `feDistantLight`
- `feSpotLight`
- `fePointLight`
- `feImage`

Is this intentional? Why?
If not, shall I send a PR adding them?

much obliged!
~Y

cure53:
I think light sources are harmless, so the three _Light_-related filters can be added. 

With _feImage_ I would be careful though, this caused XSS in the past.

ydaniv:
Yeah, that's what I thought. I'll send a PR with the light filters.

cure53:
<3 thanks!

PR #270, opened 2018-03-19
Added missing light filters in svgFilters

ydaniv:
See #269


cure53:
Looking good, thanks!

Issue #276, opened 2018-04-16
Preserve attribute order


mayfield:
> This issue proposes a feature which preserves attribute order.

I compare input and output html passed through dompurify to detect if something has changed (and thus if I need to mutate my document).  A simple string comparison generally works except when there are multiple attributes on input html.  The attributes get reordered by dompurify and the resulting string is different despite being effectively identical.  It would be great if the attribute order was preserved so simple string compares more accurately elucidated change state.

cure53:
Heya,

> The attributes get reordered by dompurify

Nope, it's the browsers - not us. There is little to nothing we can do here. You would have to reorder them manually - and reverse the browser's behavior.

mayfield:
Bummer.   Thanks for quick reply.

mayfield:
This might be impl specific but it looks like `_sanitizeAttributes`  is iterating over the attributes in reverse order and calling `removeAttribute` for every attribute regardless of sanitization being required or not.  Cleaned values are then reapplied using `setAttribute`.  According to MDN, `setAttribute` will "replace" the attribute if it already exists.  Testing in chromium seems to validate that this DOES preserve order.

Again, I don't know if that's impl specific, but for at least chrome it appears the procedure could be massaged to not call removeAttribute unless it needs to.  Additional coverage for explicitly calling removeAttribute would be required;  Such as the `setAttribute`'s `catch()` clause.

An alternative approach would be to collect the attribute map first, remove all of them from the node and then iterate forward.  Again, this is perhaps impl specific, but in chrome setAttribute appears to preserve insertion order in the resulting HTML string.

Thoughts?

mayfield:
I was able to prototype a patch that implements the latter technique and it appears to preserve attr order on chromium.  It does involving cloning the node first so the attributes NamedNodeMap order is preserved during forward iteration (and subsequent alteration).  There surely exists a technique to avoid this clone but it would require more substantial alteration.

cure53:
Nope. That won't work. We have to work through attributes from bottom to top because otherwise the index of the attributes gets screwed up and we have bypasses.

mayfield:
Re: Nope;   Did you notice the cloneNode().attributes assignment for the iteration?

PR #277, opened 2018-04-16
Attempt attribute order preservation (#2).

mayfield:
This mostly works except in cases where Safari workarounds
for img elements with id and name are set.  Perhaps that workaround
can be made more surgical for affected engines?

Tested on Chrome 65 (linux), Firefox 59 (linux) and Safari 11.0.3 (macos).

> This pull request implements #276


cure53:
> Tested on Chrome 65 (linux), Firefox 59 (linux) and Safari 11.0.3 (macos).

The thing completely blows up in Edge and MSIE :D

mayfield:
I can't infer anything about IE and Edge from travis CI, perhaps you tried those browsers by hand.  However running `npm run test:ci` did reveal a bunch of false positives due to the attribute ordering being unreversed.  I can't dedicate the time required to see this through unfortunately.  So I'm closing the PR.

PR #282, opened 2018-06-05
Add support for 'use' tag with SVGs

sjalgeo:
> This pull request [implements, fixes, changes]...

Allow usage of the `use` tag in SVG.

### Background & Context

Unsure if there's a criteria for what is included or not, please let me know if there's a reason this can't be added.

https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use

### Tasks

na

### Dependencies

na

### DIFF
--- a/src/tags.js
+++ b/src/tags.js
@@ -161,6 +161,7 @@ export const svg = [
   'title',
   'tref',
   'tspan',
+  'use',
   'video',
   'view',
   'vkern',


cure53:
`<use>` can introduce XSS. That is why we don't white-list it.

https://insert-script.blogspot.com/2014/02/svg-fun-time-firefox-svg-vector.html

Issue #283, opened 2018-06-12
Does SAFE_FOR_JQUERY work as intended here?


thorn0:
> This issue proposes a bug (probably) which...

### Background & Context

Does the `SAFE_FOR_JQUERY` mode work as intended here?
Can I make DOMPurify not change this string and stay jQuery-safe at that? Why does it do this?

```js
var html = `<i>&amp;amp; &lt;</i>`;
console.log(html === DOMPurify.sanitize(html, { SAFE_FOR_JQUERY: true })); // false!
console.log(DOMPurify.sanitize(html, { SAFE_FOR_JQUERY: true })); // "<i>&amp; &lt;</i>"
```

If everything is alright and works as intended, how exactly is the output jQuery-safer than the input? Where can I read up on this topic?

Chrome Version 67.0.3396.79 (Official Build) (64-bit), Windows 7

#### Input

```html
<i>&amp;amp; &lt;</i>
```

#### Given output

```html
<i>&amp; &lt;</i>
```

#### Expected output

Same as the input.

cure53:
Interesting, we'll have a look!

cure53:
Okay, I think I have a working fix, let's see what the tests will say :)

cure53:
The tests are green, I think we are good to close this one.

thorn0:
Seems like it should make sense to add this case to tests too.

cure53:
A PR would be very much appreciated :)

PR #284, opened 2018-06-13
Add a test for #283

thorn0:
See #283


cure53:
Awesome, thank you very much :)

Issue #288, opened 2018-06-28
FR: In-place sanitization for nodes


choumx:
This issue proposes a feature which sanitizes nodes in-place instead of importing it into a new Document and returning a sanitized copy.

Sample PR: #289

### Background & Context

`DOMPurify.sanitize(node, {RETURN_DOM: true})` returns a sanitized copy of `node`. For some use cases, it would be useful if DOMPurify sanitized `node` itself instead. 

### Feature

One idea is to add a new configuration option called `IN_PLACE` that defaults to `false`. I.e.

`DOMPurify.sanitize(node, {IN_PLACE: true})`

This would be useful for incremental sanitization use cases (e.g. DOM streaming/mutation APIs) by avoiding the performance overhead of (1) deep importing/copying `node` and (2) DOM mutation caused by replacing `node` with its sanitized copy.

The specific use case we're interested in is https://github.com/ampproject/amphtml/issues/13471. We're experimenting with running untrusted JS (e.g. a React app) in a web worker containing a virtual DOM. The virtual DOM would listen for mutations and pipe changes back to the main page. We need to sanitize those mutations to prevent DOM XSS and are interested in trying out DOMPurify to do this. We recently started using DOMPurify for runtime template sanitization -- thank you!

Any thoughts or feedback would be appreciated. Also curious whether there are security implications here.

choumx:
FYI with a quick a dirty test on TodoMVC, it looks like doing sanitization in-place offers substantial performance wins. For sanitizing the following DOM snippet (a "todo" list item):

```html
<li class="">
  <div class="view">
    <input type="checkbox" class="toggle"><label>a</label>
    <button class="destroy"></button>
  </div>
</li>
```

Out-of-place: 5-6ms
In-place: 1-2ms

cure53:
Hey there, this sounds very interesting! We would be up for it, the performance benefits sound delicious.

However, I am interested in how you got the test data on TodoMVC. Is there a PR already we could have a look at? How did you perform the mentioned benchmark?

One of the core principles of DOMPurify is not to trust the surrounding page too much and create clean copies of important properties and methods before we even touch the user controlled markup. Is this still, something we could be able to follow with this approach?

choumx:
Sorry for the delay. I was testing our prototype of the referenced feature on my workstation with 6x CPU throttling on Chrome. Here's an isolated lab test I quickly whipped up:

```html
<!doctype html>
<html>
  <head>
[...]

PR #289, opened 2018-07-03
Sample implementation for IN_PLACE

choumx:
This pull request is a prototype for in-place sanitization.

### Background & Context

Sample implementation for #288.


cure53:
I see no issues here and neither do the tests. Ready for a merge?

Also, thanks :)

choumx:
That would be grand. Thank you!

PR #295, opened 2018-08-09
Fix missing attribute namespace

choumx:
This pull request fixes an issue where attribute namespaces are not preserved. This can cause bugs when using `RETURN_DOM`, e.g. browsers that fail to render SVGs that use `<image xlink:href="...">`.

Simple test case:
```html
<svg width=50 height=50>
  <g>
    <image xlink:href="foo.svg" width=50 height=50></image>
  </g>
</svg>
```


choumx:
Thanks for the quick turnaround!

PR #303, opened 2018-09-28
Hardened to resist interference by naive code in other modules.

mikesamuel:
This doesn't make it impossible to pull the rug out from under
purify code, but does mean that it preserves semantics in the face
of naive code that does the equivalent of

*  `import(a)[b] = c` where (a, b, c) are attacker controlled strings
*  `Builtin.prototype.newProperty = polyfill`

This does not attempt to address the much larger number of
dependencies on builtin prototypes like String.prototype.match or
Array.prototype.push.

So if an application makes existing properties on the builtin prototypes
non-configurable and loads purify early then purify should preserve
semantics.

----

Specific changes include:

*  Modifies util.js addToSet to separate "set" objects from Object.prototype
*  Modifies util.js addToSet to not reassign to arrays when no change is
   needed so we can freeze whitelists like those in attrs.js and tags.js
*  Seals RegExps in regexp.js so that we can rely on their test and exec
   methods being the ones defined on RegExp.prototype.
*  Throughout, grab static methods like Object.freeze early instead of
   relying on polyfills not to overly aggressively replace them.
*  Throughout, prefers Reflect.apply to Function.prototype.{call,apply}
   which can be masked.


cure53:
Fantastic, I see no reason why not to merge this :) Thanks!

src/utils.js
```diff
@@ -1,11 +1,31 @@
...
+  }
   let l = array.length;
   while (l--) {
-    if (typeof array[l] === 'string') {
-      array[l] = array[l].toLowerCase();
+    let element = array[l];
+    if (typeof element === 'string') {
+      const lcElement = element.toLowerCase();
+      if (lcElement !== element) {
+        array[l] = lcElement;
```
choumx: @mikesamuel I believe this is problematic when passing the newly-frozen config presets into `array`: #314

PR #304, opened 2018-09-30
Add missing var declarations

styfle:
This pull request fixes a couple missing variable declarations.
Logically, there should be no change.

### Background & Context

This is really just a code cleanup.

### Tasks

I'm not sure?

### Dependencies

None


cure53:
Oooh, we missed these, thanks :D

PR #306, opened 2018-10-09
Squashing PRV Memory Leak

tlau88:
### Background & Context

Variable persists and points to a possibly large data that is not eligible for garbage collection. There possibly might be other variables eligible for assigning.

### Tasks

- Add tasks to give a quick overview for QA and reviewers

### Dependencies

If there are any dependencies on PRs or API work then list them here.

- [x] Resolved dependency
- [ ] Open dependency


cure53:
Thanks, very nice find!

PR #309, opened 2018-10-28
fix misspellings of some tags and attrs in the allowlists

flavorjones:
> This pull request fixes the spelling of several tags and attributes in the allowlists.

### Background & Context

Sanitization won't happen correctly if the tags and attributes aren't spelled correctly! :-D

### Tasks

Please feel free to double-check my corrected spellings in the `src` files against the SVG and MathML specs.

### Dependencies

None.


cure53:
Fantastic spot, thanks!

PR #310, opened 2018-10-31
Added support for Trusted Types.

koto:
DOMPurify detects if TrustedTypes are available. If so, will create an inner no-op policy, with a name 'dompurify', optionally suffixed with a value specified in a data-tt-policy-suffix attribute of the script using DOMPurify. It will use that policy when interacting with the (dirty) DOM, and will return a TrustedHTML object instead of a string.

See https://github.com/WICG/trusted-types.

This enables DOMPurify to be used on documents that have enforced Trusted Types API (see e.g. http://tinyurl.com/tttpac), and in general helps migrating web applications using DOMPurify to Trusted Types.


cure53:
Wow, this looks amazing, thank you! I cannot spot any problems for now, going for a merge.

PR #315, opened 2019-01-15
Check if array is frozen in addToSet()

### LINKED ISSUE #314: addToSet() runtime error with frozen config presets
This issue proposes a bug which...

### Background & Context

Noticed this in amphtml's test suite when attempting to upgrade to 1.0.9:

```
TypeError: Cannot assign to read only property '22' of object '[object Array]'
  at addToSet (...)
```

### Bug

The bug happens on this line: https://github.com/cure53/DOMPurify/blob/b06b6b2ac8605b88e08ff36f895097fde8c9e8f5/src/purify.js#L348

`addToSet(set, array)` will attempt to write to `array` with lowercased values, but this will fail if `array` is frozen and not writable:

https://github.com/cure53/DOMPurify/blob/b06b6b2ac8605b88e08ff36f895097fde8c9e8f5/src/utils.js#L22-L26

This manifests itself only on `svgFilters`, which is frozen and contains uppercased characters that will pass the conditional above, e.g. https://github.com/cure53/DOMPurify/blob/b06b6b2ac8605b88e08ff36f895097fde8c9e8f5/src/tags.js#L175-L176

I believe this was introduced by #303 which was released in 1.0.9. 

#### Input

Any input markup with a config that has `USE_PROFILES.svgFilters == true`.

#### Given output

Runtime error: `TypeError: Cannot assign to read only property '22' of object '[object Array]'`

#### Expected output

No runtime error.

choumx:
Fixes #314. 

Also adds a test for `{USE_PROFILES: {svgFilter: true}}`. Confirmed it fails without the `Object.isFrozen` check:

```
not ok 727 - Died on test #11     at module.exports (/Users/willchou/DOMPurify/test/test-suite.js:452:9)
#     at Object.done (/Users/willchou/DOMPurify/test/jsdom-node.js:51:5)
#     at process.nextTick (/Users/willchou/DOMPurify/node_modules/jsdom/lib/jsdom.js:297:18)
#     at process._tickCallback (internal/process/next_tick.js:61:11): Cannot assign to read only property '22' of object '[object Array]', expected: undefined, got: null, test: Config-Flag tests: USE_PROFILES, module: DOMPurify in jsdom, source: TypeError: Cannot assign to read only property '22' of object '[object Array]'
#     at addToSet (/Users/willchou/DOMPurify/dist/purify.cjs.js:56:18)
#     at _parseConfig (/Users/willchou/DOMPurify/dist/purify.cjs.js:384:9)
#     at Function.DOMPurify.sanitize (/Users/willchou/DOMPurify/dist/purify.cjs.js:929:7)
#     at Object.<anonymous> (/Users/willchou/DOMPurify/test/test-suite.js:466:34)
```


cure53:
Fantastic thank you! The code was reviewed and is okay.

PR #325, opened 2019-03-13
Add support for 'target' attribute.

classix-as:
> This pull request implements support for the [attribute target](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target).

### Background & Context

We would like to specify hyperlinks to open a new window. We do not see a security problem with the target attribute, neither in its typical form in the a-tag nor in the other tags [area](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/area#attr-target), [base](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base#attr-target), [form](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-target), in which target essentially has the same semantics.

### Tasks

- Please make sure that you agree with our interpretation of the attribute target.

### Dependencies

N/A


cure53:
Hey all, you can easily add support for the `target` attribute using a configuration directive, 
such as for example:

`var clean = DOMPurify.sanitize(dirty, {ADD_ATTR: ['target']});`

We have no plans to add the `target` attribute to the default whitelist as it might aid an attacker in performing tab-nabbing attacks, pre-populating `window.name` and the likes.

PR #329, opened 2019-04-02
sanitize input[type=file] if hookEvent.keepAttr is false

alabiaga:
This pull request allows for the sanitization/blacklisting of `<input type="file" />` by setting the hookEvent keepAttr to false.

### Background & Context

Currently an [AMP HTML](https://github.com/ampproject/amphtml)  spec disallows for the following usage in templates: input[type=file]
DOM Purify does not allow for the sanitization of the type `file` because of a Safari bug which doesn't permit the adding back of this attribute once removed. See https://github.com/cure53/DOMPurify/blob/master/src/purify.js#L843


cure53:
LGTM, thank you!

PR #335, opened 2019-04-27
Added missing SVG filter attributes

ydaniv:
`filterUnits` and `primitiveUnits`. These appear to have been left out without intention.

Thanks!
~Y


cure53:
:+1: thanks!

PR #342, opened 2019-07-02
Add href as a data uri attribute

hansottowirtz:
This pull request adds `href` as a safe data uri attribute, for SVG image tags.

### Background & Context

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/href), `xlink:href` is deprecated and replaced by `href`.


samsnowy:
@cure53 Glad to see this fixed, do you know when the next npm package is going to be published? I'm currently using xlink:href as a kludge to get it working for svg image tags but I would like to use href soon.

PR #368, opened 2019-09-26
Use RegExp.test() instead of String.match()

jorangreef:
RegExp.test() returns a boolean whereas String.match() returns an array
of match results, requiring object allocation and adding GC pressure.

If match results are not needed then RegExp.test() is faster.


cure53:
Thank you :)

Issue #372, opened 2019-10-18
Crashes on IE11 if sanitizing a space


sontek:
On IE11:

```
DOMPurify.sanitize(" ", {USE_PROFILES: {html: true}})
Unable to get property 'insertBefore' of undefined or null reference
```

Using latest version (2.0.6), you can verify it on:

https://cure53.de/purify

by opening the console and pasting the line of code above.

cure53:
Oh, nice spot! We will release 2.0.7 soon and the bug will be fixed

peterblazejewicz:
@sontek 
This can be closed, as the 2.0.7 was released (and one of the product I work with uses IE11, so I've checked this thread).
Thanks!

peterblazejewicz:
Hm,
so the '2.0.7' is not released via NPM yet? 
The online version states '2.0.7' as the `DOMPurify.version`

cure53:
Will be released very very soon :)

cure53:
Pretty much on it right now

cure53:
Alright, this fix should address the problem you described:

https://github.com/cure53/DOMPurify/commit/56b9e9caa85ee61664c86ba16306681ed9d6dbb0

Does it work as intended?

Issue #373, opened 2019-10-21
Whitespace sanitization behavior changed


makepanic:
> This issue proposes a bug which...

### Background & Context

Hi,
thanks for developing this great package.
I'm in the progress of updating from 1.x.x to 2.x.x and noticed our tests failing.

It seems like 2.x.x introduced a changed behavior when sanitizing a whitespace character string.
Previously it returned the whitespace string but now it's an empty string.

I'm not sure if this is a bug or expected behavior. We're invoking the `sanitize` method like this: `sanitize(text, { ALLOWED_TAGS: [] })`

#### Input

` ` - (whitespace character ASCII 32)

#### Given output

`` - (empty string: string.length === 0)

#### Expected output

` ` - (whitespace character ASCII 32)

peterblazejewicz:
@makepanic 
what is the version you're using?
See: #372

cure53:
I think I found the problem. 

MSIE11 won't create a full document if `body.outerHTML` gets set to an empty string. That would cause the exception. With the fix, we addressed #372 successfully but indeed the root cause was something else.

cure53:
Let's see what the other browsers say... :D

cure53:
I think master is ready for a quick review, should be fixed imho.

makepanic:
> what is the version you're using?

2.0.7 tested it only in chrome

makepanic:
Just tried `cure53/DOMPurify#3a303715d73de342e25e8437cc59552334823092` in our tests and it seems to have fixed the whitespace regression :+1:

cure53:
Awesome :D

PR #374, opened 2019-10-21
:white_check_mark: test case for #372

peterblazejewicz:
Add test case for sanitizing white space character returning empty
string. This bounds the behaviour with test case.

See: #372

### DIFF
--- a/test/test-suite.js
+++ b/test/test-suite.js
@@ -520,6 +520,7 @@ module.exports = function(DOMPurify, window, tests, xssTests) {
           '<svg keep="me"></svg>', 
           "<svg xmlns=\"http://www.w3.org/2000/svg\" keep=\"me\" />"
       ] );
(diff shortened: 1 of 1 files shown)


cure53:
Hmm, does this not collide with #373?

cure53:
I think this is what we want, no?

https://github.com/cure53/DOMPurify/commit/3a303715d73de342e25e8437cc59552334823092

Fixes the crash and restores correct white-space behavior.

peterblazejewicz:
Nope, I've added test for a behaviour we discussed today, exhibited on the website version. Now I see that's no longer relevant.
The test should probably cover the another use case with longer white space string (so it proves no trimming behaviour happens, as it was before).
Closing this one,  thanks for the head ups!

cure53:
Ah, all good then :) Thank you!

PR #378, opened 2019-12-20
Support the <picture> element

chris-morgan:
The `<picture>` element is a useful extension over the `` element, allowing things like selecting a different source for an image if `(prefers-color-scheme: dark)` (useful for emails), without having to resort to a background-image or more convoluted structure.

Before this, any `<source>` or `` elements inside a `<picture>` would be retained, but the containing `<picture>` element would be removed, leaving just an orphan `<source>` element, and the fallback ``—which is sane behaviour and the intent of the `<picture>` element, but doesn’t let you *use* it.

Individual users could specify ALLOWED_TAGS and there add `'picture'`, but that’s a pain to maintain.

I see no hazard in including this element for everyone. The element itself has no meaningful surface, so any increase in surface is in the child `<source>` element which will now have meaning rather than being malformed, most notably with its `srcset` attribute, which `` can already have, so I don’t believe it’s any new surface.


chris-morgan:
CI failure is spurious.

cure53:
Looks good, makes sense, thanks :)

PR #385, opened 2020-01-15
More conservative uponSanitizeAttr.forceKeepAttr = true

devinrhode2:
This is a much more conservative version of my previous PR: https://github.com/cure53/DOMPurify/pull/384

I realized that I was changing default behavior, and potentially disabling removing any and all attrs


cure53:
This looks, great, thanks!

Issue #388, opened 2020-01-30
Difficulty removing a node


mikesnare:
See https://jsbin.com/kokuqecoci/8/edit?js,console for a working example of the issue.

I have some node types that I need fully removed -- i.e. not just the tag (leaving the content) but the node and all its children.  I have some node types where I need the tag stripped but the content left in place, and I have some nodes where the node and it's content are fine, subject to the previous rules.

As a simplified example, say that `DIV`s are fully allowed, but `SPAN`s need to be completely removed, including content.

So, given the input `<div><span>This is my span</span></div>` what I need for output would be `<div></div>` (this is simplified example to demonstrate the problem).

In order to achieve this I set up the following:

```js
var CONFIG = {
  ALLOWED_TAGS: [
    'div',
    '#text'
  ],
  ALLOWED_ATTR: [],
  KEEP_CONTENT: true,
};

window.DOMPurify.setConfig(CONFIG);

DOMPurify.addHook('uponSanitizeElement', function(node) {
  console.log('uponSanitizeElement: ' + node.nodeName);
  if (node.nodeName === 'SPAN') {
    console.log('Attempting to remove ' + node.nodeName + ' from ' + node.parentNode.nodeName);
    return node.parentNode.removeChild(node);
  }
});


try {
  var html = "<div><span>This is my span</span></div>";
  console.log(DOMPurify.sanitize(html).trim())
} catch(e) {
  console.error(e);
}
```

When I run this, I get errors in the console due to the removal of the SPAN.  In the demos folder, the example for removing nodes uses the `beforeSanitizeAttributes` hook, but that hook does not get called back for me for the `span` in this example.

I'm perfectly willing to believe that I'm just doing something wrong, but I don't know what that might be.  Any help would be appreciated.

cure53:
Heya, 

hmmm - have you tried to add the SPAN`and others but DIV to `FORBID_CONTENTS`?
https://github.com/cure53/DOMPurify/blob/master/src/purify.js#L253

mikesnare:
Didn't know that was a thing, so thanks!  Looks like it does what I need without any callbacks needed, but it's also a private internal config so I can't change it without modifying the source.  Would you accept a PR to fix that?

mikesnare:
[...]

PR #389, opened 2020-01-31
Adding a new example for node removal

mikesnare:
This PR adds a demo of removing nodes when you only want to allow a subset of tags, which is not covered by the existing node removal demo.

### Background & Context

See https://github.com/cure53/DOMPurify/issues/388

### Tasks

Just open the new demo page in the browser.

### Dependencies

There are no dependencies


cure53:
This is awesome, thank you very much! :)

PR #393, opened 2020-02-28
Updating build to inject license text at start of output files

mscheele7:
> This pull request adds attribution comments at the start of the distributed artifacts

### Background & Context

Updated build process to inject the license comment into output files and prevent uglify from removing it.


mscheele7:
Missing item(s) on my end, will reopen/resubmit after fixing.

PR #394, opened 2020-03-02
Update build to inject license text at start of artifacts

mscheele7:
> This pull request adds attribution comments at the start of the distributed artifacts

### Background & Context

Updated build process to inject the license comment into output files and prevent uglify from removing it.


cure53:
This looks okay, thanks. 
However, should we not make use of a smaller license block? Or is the full thing needed?

My understanding was that we only need a one-liner. Something like:

`/*! DOMPurify | (c) Cure53 and other contributors | github.com/cure53/DOMPurify/blob/master/LICENSE */`

mscheele7:
That was my initial understanding, but that larger block is what I was told to use.

cure53:
Any way we can reduce that footprint a but?

cure53:
In other words, we should reduce this to a one liner, this way it is too much overhead and I can already see incoming requests from folks all over the place to roll that back.

Too much of a footprint is a blocker for us.

mscheele7:
I'm being told that a shorter comment is acceptable, as long as I'm not the one to shorten it.

Also that apparently the Apache license doesn't technically allow for a shorter comment (skimming through the stuff with short one-liners a lot are MIT or such).

cure53:
Gotcha, I will merge and then make the file a one-liner.

PR #396, opened 2020-03-12
Update uglify configuration to preserve license comment in minified JS

mscheele7:
> This pull request fixes the updated license comment not getting included in the minified JS

### Background & Context

The prior long license comment used a license annotation to signal Uglify to preserve it with the comments: 'some' setting. Since that annotation is no longer present this changes the setting to use a regex to match the comment's leading !.


cure53:
LGTM, thx! :+1:

PR #407, opened 2020-03-24
Update license header to point to specific version of LICENSE

mscheele7:
### Background & Context

Legal raised concerns that the URL in the license header pointed to master rather than a specific commit. This updates it to point to a release version, as well as name the specific licenses.


mscheele7:
Sorry, should have been more clear.

The concern is that if a future version uses a different license the LICENSE file will change. Then any previous releases will be pointing to the new license file instead of the license those versions were released under.

cure53:
Gotcha, that makes sense!

PR #418, opened 2020-04-18
refactor: to update tooling (rollup, babel et al) to recent versions

tdeekens:
> This pull request updates all dependencies related to tooling to their recent versions.

### Background & Context

The versions were quite outdated. Before we proceed to take on CI and maybe other things I suggest we get our tooling stack up to date. This includes:

1. Babel (it's all scoped packages by now)
2. Rollup (moved quite quickly and we picked the right bundler two years back)
3. Prettier chores the code a bit but that's the nature of formatters
4. ESLint and XO changes a bit. I disabled some rules I didn't want to use in code. Some rather okay suggestions I changed
5. Suggesting to use terser instead of uglify to minify our code


.nvmrc
```diff
@@ -0,0 +1 @@
+13
```
tdeekens, line 1: Just helps when working locally that we all use the same node version. Plus we can use it later on CI.

rollup.config.js
```diff
@@ -1,44 +1,37 @@
-const nodeResolve = require("rollup-plugin-node-resolve");
-const babel = require("rollup-plugin-babel");
-const replace = require("rollup-plugin-replace");
-const uglify = require("rollup-plugin-uglify");
-const fs = require("fs");
+const nodeResolve = require('rollup-plugin-node-resolve');
+const babel = require('rollup-plugin-babel');
+const replace = require('rollup-plugin-replace');
+const { terser } = require('rollup-plugin-terser');
```
tdeekens, line 4: I know this is a sensitive topic here and I am not forcing this. I just wanted to bring up the suggestion. Terser is quite happily adopted in the community by now. The main reason is that uglify didn't move to support newer language features if I recall.

1. https://github.com/terser/terser
2. https://github.com/mishoo/UglifyJS2

I am fine either way but didn't have any bad experiences with terser.

src/purify.js
```diff
@@ -947,7 +949,7 @@ function createDOMPurify(window = getGlobal()) {
         }
 
         arrayPop(DOMPurify.removed);
-      } catch (error) {}
+      } catch (_) {}
```
tdeekens, line 952: This is ESLint fixes. Cause we don't use the arg to indicate that better.

test/karma.conf.js
```diff
@@ -1,4 +1,4 @@
-const commonjs = require("rollup-plugin-commonjs");
+const commonjs require('@rollup/plugin-commonjs');
```
tdeekens, line 1: The other plugin is deprecated.

tdeekens:
Empty TravisCI logs. What’s up there. Lets see another day.

tdeekens:
The main problem here is that Node.js moved long on and is on v13 now [ref](https://nodejs.org/en/about/releases/).

CleanShot 2020-04-19 at 10 01 49@2x

That doesn't mean we can't support these older versions but unlikely can test on them cause all dependencies have moved on. Meaning fsevents or cross-env doesn't support the old versions anymore.

Maybe we can note something in the readme at a later state which versions we actively support and which ones we can't guarantee support for?

tdeekens:
Yay, at least now it's a test failing. Maybe @cure53 you have an idea. That's something again I know not too much about. Do you see any change here that could cause this?

rollup.config.js
cure53: :+1: Fine with trying terser

src/purify.js
cure53: Yeah, had to do the same a while ago

cure53:
> The main problem here is that Node.js moved long on and is on v13 now [ref](https://nodejs.org/en/about/releases/).
> 
> CleanShot 2020-04-19 at 10 01 49@2x
> 
> That doesn't mean we can't support these older versions but unlikely can test on them cause all dependencies have moved on. Meaning fsevents or cross-env doesn't support the old versions anymore.
> 
> Maybe we can note something in the readme at a later state which versions we actively support and which ones we can't guarantee support for?

I think, just as with older browsers, we need to phase out support for older versions. Happy to draft a text for the README once we have clarity with versions will be supported.

tdeekens:
My goal would be to get GitHub Actions in. Then rebase this ontop master and then move on with this.

tdeekens:
> I think, just as with older browsers, we need to phase out support for older versions. Happy to draft a text for the README once we have clarity with versions will be supported.

Yeah I agree. I think bottom line it's that we don't break with older Node.js versions yet and they continue to work. We however don't test for them anymore. Implying that potentially regressions will not be spotted from now on (if we merge these updates here). In future terms it would also mean that we similarily follow the Node.js release cycle and support any active version. However, phase out support for any version entering maintainance. As we do now with this.

cure53:
> Yeah I agree. I think bottom line it's that we don't break with older Node.js versions yet and they continue to work. We however don't test for them anymore.

This makes perfect sense!

cure53:
Is this one ready for merging as well?

tdeekens:
Bloody hell. This is more fun than I thought. Seems like no dependency moved for 3 years. I reduced it to a meaningful update. From here we can move stepwise if possible and worthwhile for the lib :)

tdeekens:
Yes, I will merge this once it passes.

README.md
```diff
@@ -40,11 +40,11 @@ Afterwards you can sanitize strings by executing the following code:
 var clean = DOMPurify.sanitize(dirty);
 ```
 
-The resulting HTML can be written into a DOM element using `innerHTML` or the DOM using `document.write()`. That is fully up to you. But keep in mind, if you use the sanitized HTML with jQuery's very insecure `elm.html()` method, then the `SAFE_FOR_JQUERY` flag has to be set to make sure it's safe! Other than that, all is fine. 
+The resulting HTML can be written into a DOM element using `innerHTML` or the DOM using `document.write()`. That is fully up to you. But keep in mind, if you use the sanitized HTML with jQuery's very insecure `elm.html()` method, then the `SAFE_FOR_JQUERY` flag has to be set to make sure it's safe! Other than that, all is fine.
```
tdeekens, line 43: Just auto-formatting. Will follow-up with an npm run script for it.

test/test-suite.js
```diff
@@ -1,668 +1,1453 @@
-module.exports = function(DOMPurify, window, tests, xssTests) {
+module.exports = function (DOMPurify, window, tests, xssTests) {
```
tdeekens, line 1: Oh sorry. Didn't intend to change this. Will revert it to master.

PR #419, opened 2020-04-19
feat: add github actions workflow and removes travis

tdeekens:
> This pull request adds a GitHub Action to perform the tests.

### Background & Context

The workflow is similar to what we have ton TravisCI. It removes some node versions and has the same semantics of probe tests and testing against BrowserStack.


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,50 @@
...
+        run: yarn install --frozen-lockfile
+      - name: Build
+        run: npm run build
+      - name: Lint
+        run: npm run lint
+      - name: Test
+        env:
+          TEST_BROWSERSTACK: ${{ startsWith(matrix.node-version, '13') }}
+          TEST_PROBE_ONLY: ${{ github.ref != 'refs/heads/master' }}
+          BS_USERNAME: ${{ secrets.BS_USERNAME }}
```
tdeekens, line 48: @cure53 I don't see the `Settings`-tab under the repository. Can you add these as secrets on GitHub. You can find them here. I suppose we have to generate a new set on BrowserStack for the GitHub Actions. Still one we got "rid of" TravisCI we can remove the old ones.

CleanShot 2020-04-19 at 11 08 53@2x

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,50 @@
...
+          node-version: ${{ matrix.node-version }}
+      - name: Install
+        run: yarn install --frozen-lockfile
+      - name: Build
+        run: npm run build
+      - name: Lint
+        run: npm run lint
+      - name: Test
+        env:
+          TEST_BROWSERSTACK: ${{ startsWith(matrix.node-version, '13') }}
```
tdeekens, line 46: For this I like GitHub Actions.

package.json
```diff
@@ -13,7 +13,7 @@
     "build:cjs": "cross-env NODE_ENV=development BABEL_ENV=rollup rollup -c -f cjs -o dist/purify.cjs.js",
     "test:jsdom": "cross-env NODE_ENV=test BABEL_ENV=rollup node test/jsdom-node-runner --dot",
     "test:karma": "cross-env NODE_ENV=test BABEL_ENV=rollup karma start test/karma.conf.js --log-level warn ",
-    "test:ci": "cross-env NODE_ENV=test BABEL_ENV=rollup npm run lint && npm run test:jsdom && npm run test:karma -- --log-level error --reporters dots --single-run --shouldTestOnBrowserStack=\"${TEST_BROWSERSTACK}\" --shouldProbeOnly=\"${TRAVIS_PULL_REQUEST}\"",
+    "test:ci": "cross-env NODE_ENV=test BABEL_ENV=rollup npm run test:jsdom && npm run test:karma -- --log-level error --reporters dots --single-run --shouldTestOnBrowserStack=\"${TEST_BROWSERSTACK}\" --shouldProbeOnly=\"${TEST_PROBE}\"",
```
tdeekens, line 16: I split out the lint. I gives them as two steps on CI.

tdeekens:
Odd. Failing with the same test as on my other PR. What Am I doing wrong. I branched off master. Hm.

CleanShot 2020-04-19 at 11 11 07@2x

.github/workflows/build-and-test.yml
cure53: What values to initialize them with?

cure53:
> Odd. Failing with the same test as on my other PR. What Am I doing wrong. I branched off master. Hm.
> 
> CleanShot 2020-04-19 at 11 11 07@2x

This might be a temporary artifact. will look into this once we got the rest working

.github/workflows/build-and-test.yml
cure53: Added them, two set to `true` and one to the username accordingly

tdeekens:
It seems CI (GitHub or Travis) independent while there is no code change made. Maybe I‘ll just try again in a couple of days. Is that what you are saying? 

My plan was:

- GitHub Actions over Travis
- Dependency house keeping
- Documentation improvements in readme
- Tinker with CodeSandbox to provide examples

Would that work for you? It would also imply that we should be able to try to proceed with this first. I wanted to check to get the same browser setup as on Travis. From there we only need the secrets to my knowledge. 

Thanks for the swift response 😎🤘🏼

tdeekens:
@cure53 let me know when you added the secrets. 🤗

cure53:
@tdeekens Just to avoid me doing something stupid, you need me to add the BS_ACCESSKEY value as secret, correct?

tdeekens:
Yes. In the GitHub settings under secrets we need both referenced secrets referenced by the GitHub Action yml. They should be added in clear text. They are never exposed from the UI and can not be read (only deleted) once added. It is a bit different to TravisCI where you commit the encrypted value but most CIs I have worked with recently go with the UI based solution. Sorry, long story: just add it in the project‘s settings as a secret. 😎

cure53:
Done :)

tdeekens:
Cool! What do we do about the failing test here? I remain clueless. Any hint or help would be ace 🐼.

cure53:
> Cool! What do we do about the failing test here? I remain clueless. Any hint or help would be ace panda_face.

As far as I can see, nothing security critical, we can fix that later.

tdeekens:
> As far as I can see, nothing security critical, we can fix that later.

Okay but what are you suggesting. I don't want to merge anything with failing CI 😄. 

Let me try to fix it to see if the BrowserStack integration would work.

tdeekens:
Alright, there we do. This does the trick now. We could take the gamble and merge it and say 👋 TravisCI 😄.

cure53:
Mergiiing! :D

PR #422, opened 2020-04-19
chore: apply prettier to more

tdeekens:
> This pull request applies prettier to more things.

### Background & Context

I noticed that we don't auto-format as much as we could. Just a suggestion to broaden it a bit.


tdeekens:
Last one for today. Promise :)

PR #429, opened 2020-04-22
Remove yarn error logfile and it to the .gitignore

LeSuisse:
### Background & Context

It seems to have been initially committed by mistake in c9661e1a4c2e32b0ad87807b12eb33edc1361095 as it serves no purpose in the repository.
    
As this file contains information about the local environment of the developer and to avoid to commit it again by mistake, the file has been added to the .gitignore.


tdeekens:
Thanks!

Issue #427, opened 2020-04-22
2.0.9 overrides adds a node binary to the PATH in npx calls


dustym:
The 2.0.9 point release installs a `node` binary in the `node_modules/.bin`. In the event that your project adds `node_modules/.bin` to the current PATH, that `node` will take precedence over the system node. In my case I use `npx` which does this PATH management automatically.

This is due to the [commit](https://github.com/cure53/DOMPurify/commit/b26420eef1b10cb8116ca1b774da41896a930d34) that adds the `node` npm package as a dependency. Is it an absolute requirement or can it be installed as a dev dependency?

Example of this unexpected behavior:

```
~/dev/testpackage
❯ npm install dompurify@2.0.7
+ dompurify@2.0.7
added 1 package from 1 contributor and audited 1 package in 0.762s
found 0 vulnerabilities


~/dev/testpackage took 2s
❯ npx node -v
v10.18.0

~/dev/testpackage
❯ npm install dompurify@2.0.9

> node@13.13.0 preinstall /Users/dustym/dev/testpackage/node_modules/node
> node installArchSpecificPackage

+ node-darwin-x64@13.13.0
added 1 package in 5.234s
found 0 vulnerabilities

+ dompurify@2.0.9
added 3 packages from 2 contributors, updated 1 package and audited 4 packages in 7.376s
found 0 vulnerabilities


~/dev/testpackage took 8s
❯ npx node -v
v13.13.0
```

tdeekens:
Thanks for reporting. Do you know which dependency installs node locally?

adamdecaf:
I think this is causing https://github.com/Redocly/redoc/issues/1242 in the redoc project as well.

milesrichardson:
As an additional datapoint, this is also causing install to fail when using yarn 2 ("berry") with nodeModules linker. The downstream dependency in my case is `swagger-ui-react`:

```
➤ YN0000: ┌ Link step
➤ YN0007: │ node@npm:13.13.0 must be built because it never did before or the last one failed
➤ YN0009: │ node@npm:13.13.0 couldn't be built successfully (exit code 1, logs can be found here: /tmp/xfs-2c48a784/build.log)
➤ YN0009: │ node@npm:13.13.0 couldn't be built successfully (exit code 1, logs can be found here: /tmp/xfs-2c48a784/build.log)
➤ YN0000: └ Completed in 12.34s
➤ YN0000: Failed with errors in 58.33s

[...]

PR #432, opened 2020-04-23
fix: remove node and nvm

tdeekens:
To remove this.


LeSuisse:
I think there has been some confusion in #427. AFAIK npm/yarn does not pull dev dep of subpackages, so keeping them would probably been fine.

https://docs.npmjs.com/files/package.json#devdependencies

millette:
@LeSuisse sure - but why would node or nvm be added to (dev)dependendies at all?

* [nvm](https://www.npmjs.com/package/nvm) clearly states: "This is NOT the correct nvm. Visit http://nvm.sh and use the curl command to install it."
* [node](https://www.npmjs.com/package/node) might help a little tiny bit for dev, but it's not how it's generally done.

We'll know soon enough if anything breaks since there was a new release.

@cure53 might be a good idea to deprecate 2.0.9

```sh
npm deprecate dompurify@2.0.9 "Buggy release, please skip." # or something to that effect
```

PR #444, opened 2020-06-08
Avoid number literals for nodeTypes - fixes #441

mozfreddyb:
This pull request fixes #441

### Background & Context

simply replacing literals with Node.FOO constants.

### Tasks

- Use https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType#Constants or browser developer tools to compare numbers with names.

PR #448, opened 2020-06-22
Avoid number literals for nodeTypes - fixes #441

### LINKED ISSUE #441: use nodeType constants defined in e.g., Node.TEXT_NODE instead of hard-coded number lieterals
This issue proposes a tiny code enhancement that favors constants defined in the DOM standards over hard-coded numbers. E.g., comparing a node's `nodeType` with `Node.TEXT_NODE` rather than the number `3`. This is mostly for readability.

### Background & Context

Code like 
```js
      /* Fix IE's strange behavior with manipulated textNodes #89 */
      if (currentNode.nodeType === 3 && currentNode === oldNode) {
        continue;
      }```

or 
```js  if (!window || !window.document || window.document.nodeType !== 9) {`

would become more readable.

mozfreddyb:
This pull request fixes #441 and supsersedes #444 

### Background & Context

simply replacing literals with window.Node.FOO constants.

### Tasks

- Use https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType#Constants or browser developer tools to compare numbers with names.


mozfreddyb:
The commit hook keeps overwriting things like "window.Node.DOCUMENT_NODE" into "Node.DOCUMENT_NODE" and I don't want to skip using the commit hook. Grumble.

PR #463, opened 2020-08-31
docs: Fix simple typo, retrun -> return

timgates42:
There is a small typo in dist/purify.cjs.js, dist/purify.js, src/purify.js.

Should read `return` rather than `retrun`.


cure53:
Nice catch, thanks!

PR #465, opened 2020-09-03
Don't remove comment tags during sanitization

NicholasEllul:
This Pull Request adds the `<!-- Comment Tag -->` to the list of safe tags within DOMPurify so that it is not removed during sanitization.

### Background & Context

Right now DOMPurify strips away any comments within the user's input. You can test this by entering a comment such as `<!-- Hello World -->`into https://cure53.de/purify and seeing that it is not included in the output. 

The reason for this is because of [this block of code](https://github.com/NicholasEllul/DOMPurify/blob/066a8b347ee57eb6bb120af2f6013a0cad9a3ea5/src/purify.js#L684-L705) below. When we see that a comment isn't in our allowed tags list, we eventually force remove it which strips it from our output.

```js
    /* Remove element if anything forbids its presence */
    if (!ALLOWED_TAGS[tagName] || FORBID_TAGS[tagName]) {
       ...
       ...
       ...

      _forceRemove(currentNode);
      return true;
    }
```

I don't believe there is any security risk in allowing HTML comments to pass through the sanitizer by default. However, please correct me if i'm wrong and instead I can add a configuration option to allow comments instead.

### Tasks

- Verify that any input containing comments still preserves comments in DOMPurify's output


cure53:
Heya, thanks for the PR! However, comments sadly indeed carry a bit of a security risk, i.e. in classic Angular, they are used to mark where data from `ng-include` would be rendered etc.

Further, several existing XSS vectors make use of flaws caused by different comment parsing behaviors on client and server, see here for examples: http://html5sec.org/?comment

We would prefer not to have comments be enabled by default :)

cure53:
> However, please correct me if i'm wrong and instead I can add a configuration option to allow comments instead.

I think this would be the way to go indeed :)

NicholasEllul:
Sounds good, i'll modify this PR. Thanks for the TIL and quick response!

cure53:
Most welcome and thank you!

cure53:
By the way, one thing I am wondering, would this not work to address your issue?

```html
<!doctype html>
<html>
  <body>
    <script type="text/javascript" src="../purify.js"></script>
    <script>
    
        let rawHTML = `<s>hello</s><!-- hello  -->`;
        let sanitizedHTML = DOMPurify.sanitize(
            rawHTML, {
                ADD_TAGS: ['#comment']
        });
        console.log( sanitizedHTML );
    </script>
  </boy>
</html>
```

NicholasEllul:
Adding `#comment` to the list of `ADD_TAGS` works only if you have content that occurs before the first comment. If you attempted to do 

```js
 let rawHTML = `<!-- hello  --><s>hello</s>`;
        let sanitizedHTML = DOMPurify.sanitize(
            rawHTML, {
                ADD_TAGS: ['#comment']
        });
        console.log( sanitizedHTML );
```

Then the comment would still be removed. This is due to a bug in DOMParser where if it attempts to parse the DOM of something that begins with a comment, it seems to remove the comment. 

You can see that in this example here: 
image

I have a workaround for this im using in my project, but i'll close this PR until I find a more general workaround that would fit all DOMPurify users use-cases

cure53:
I think we have good news here as well, try the `FORCE_BODY` flag :D

PR #474, opened 2020-10-07
Remove mention of the removed SAFE_FOR_JQUERY flag in README

MatmaRex:
The flag was removed in 661571886064ee7465bcf7d295cfe5038beab9a4, other documentation about it was removed in 8177881230f019d5d7c6e607f8dbe65e28138c82.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,8 +46,6 @@ The resulting HTML can be written into a DOM element using `innerHTML` or the DO
 
 Well, please note, if you _first_ sanitize HTML and then modify it _afterwards_, you might easily **void the effects of sanitization**. If you feed the sanitized markup to another library _after_ sanitization, please be certain that the library doesn't mess around with the HTML on its own.
 
-jQuery does exactly that and that is why we have this flag mentioned above.
-
 ### Okay, makes sense, let's move on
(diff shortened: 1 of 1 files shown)


cure53:
Whoa, well spotted, thanks :D

rugk:
Can you maybe elaborate on the removal? I see no bigger docs on this change.

https://github.com/cure53/DOMPurify?tab=readme-ov-file#removed-configuration says no replacement is needed. I am anyway unsure on why? Does this depend on the jQuery version or what?

https://github.com/cure53/DOMPurify/releases/tag/2.1.0 mentions "we're safe now", but this does not depend on a specific jQuery version, does it?

cure53:
The thing is, jQuery does weird things with its own markup parser and assembled HTML in such way that HTML would be fine without jQuery, i.e. via innerHTML, but unsafe when jQuery is involved, i.e. with its `html()` method or countless others.

In the dark past, DOMPurify was not able to handle that without a config option, but our own sanitization has improved over time, so we are safe against those issues with our without the flag. So, right now it has no effect anymore and it's no longer needed.

Does that help clarify?

PR #488, opened 2020-11-16
Fix multi-license declaration

jochenberger:
See https://docs.npmjs.com/cli/v6/configuring-npm/package-json#license


cure53:
Nice, thanks!

PR #503, opened 2021-01-21
Fixed playground syntax error

PatrickSachs:
> This pull request is a quick to the playground at https://cure53.de/purify

### Background & Context

The playground currently raises a syntax error due to the training comma in the test JS file that is converted to JSON.


cure53:
Oh dear :see_no_evil: - thanks for spotting this!

Issue #507, opened 2021-01-28
(Updated Description) ALLOWED_TAGS allows tags other than those mentioned in the list


virajkulkarni14:
> This issue reports a bug which *does not* filter the tags as per a whitelist of allowed tags.

### Background & Context

I am creating a global DOMPurify config for my React app using the `setConfig` method. That config object also contains the `ALLOWED_TAGS` attribute that contains an array of HTML tags I want to restrict my content to. But the `sanitize` method still renders the unwanted tags.

### The Config file
`dompurify.config.js`:

```javascript
import { isString } from 'lodash'
import DOMPurify from 'dompurify'

// Allow only '<a></a>', '<b></b>', '<i></i>', '<p></p>' and '<u></u>' tags
const WHITELISTED_TAGS = ['a', 'b', 'i', 'p', 'u']

DOMPurify.setConfig({
  USE_PROFILES: { html: true }, // Only allow HTML
  IN_PLACE: true, // In place mode for faster sanitization,
  ALLOWED_TAGS: WHITELISTED_TAGS, // Only allow tags specified in the whitelist above
  ADD_ATTR: ['target'] // Allow elements with the target attribute
})

/**
 * Add hook to sanitize elements with a `target` attribute.
 */
DOMPurify.addHook('afterSanitizeAttributes', (node) => {
  const target = node.getAttribute('target')
  if (isString(target) && target.toLowerCase() === '_blank') {
    node.setAttribute('rel', 'noopener noreferrer')
  } else {
    node.removeAttribute('target')
  }
})

export { DOMPurify }

export default DOMPurify.sanitize
```

### The React component file
`MyComponent.js`:

```javascript
import React from 'react'
import Typography from '@material-ui/core/Typography'
import { sanitize } from 'path/to/dompurify.config'
// Other imports and unrelated code

<Typography
	component='p'
	dangerouslySetInnerHTML={{ __html: sanitize(label) }}
/>

// More unrelated code
```
#### Input

Here, inside the `dangerouslySetInnerHTML()`, if I call `sanitize()` with the below input:

`sanitize('<img src=x onerror=alert(1)//>')`

#### Given output

I get the below output:

`<img src="x">`

#### Expected output

Since I haven't included the `img` tag in the `WHITELISTED_TAGS` array in the above config file, I expected the below output, an empty string:

`''`

I am going by the below example in the project README:
```javascript
// allow only <b> elements, very strict
[...]

PR #509, opened 2021-02-02
Updated README as discussed in issue #507

virajkulkarni14:
> This pull request adds a new comment to the existing list of code examples in the README.

### Background & Context

Added a comment at the first occurrence of the `USE_PROFILES` setting to warn that it will override an existing `ALLOWED_TAGS` setting. We had discussed this in issue #507.

### Tasks

- Ensure that this is indeed the correct place to add the comment
- If not, please suggest another place to add this comment

### Additional notes

I will be creating another PR for the demos README to add an example config.


cure53:
Fantastic, thanks a lot :heart:

virajkulkarni14:
Damn, that was fast @cure53!

PR #510, opened 2021-02-02
Updated the demos README as discussed in issue #507

virajkulkarni14:
> This pull request adds a new example to the list of config demos

### Background & Context

Added a section after the advanced config example to warn that `USE_PROFILES` setting will override an existing `ALLOWED_TAGS` setting. This includes a code example that demonstrates how using both these settings will result in unexpected output. This is to supplement an existing PR #509 with a code example.

### Tasks

- Review example to see if anything is amiss
- Suggest any improvements or another place for the example to be added to

### DIFF
--- a/demos/README.md
+++ b/demos/README.md
@@ -50,6 +50,26 @@ var config = {
 var clean = DOMPurify.sanitize(dirty, config);
 ```
 
+#### A Gotcha in the Config
+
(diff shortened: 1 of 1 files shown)


cure53:
This one appears to bloat the README a bit too much - would prefer to hold this one back for now and wait if more people run into this exact problem.

Then we might reconsider.

virajkulkarni14:
Sounds good, thanks. I just thought adding a code example might help folks already running into this issue. But I agree with you that we should hold.

Also feel free to edit this PR to something shorter. Thanks again!

cure53:
Thank you!

PR #514, opened 2021-02-13
add fallback for lookup-getter util

vlad-borodaev:
> This pull request fixes an issue for lookupGetter utility.

### Background & Context

When there is no entries for properties, it returns null. Instead as a fallback the utility could return a function, which could be used i.e. to get nullable parent node.


cure53:
Hmm, looks interesting and reasonable. May I ask, what was the situation creating the need fr this PR?

vlad-borodaev:
I used an Angular wrapper for this lib (https://github.com/TinkoffCreditSystems/ng-dompurify) in Angular Universal project. It might be wrong but the issue seems to be caught when using SSR, because I haven't got such an issue like "getParentNode is not a function" when running Angular the common way

cure53:
I see, thanks. Let's see what the tests will say once merged :)

PR #527, opened 2021-03-23
Adding missing `slot` attribute

moefinley:
### Background & Context

Currently content within custom elements will have their `slot` attribute removed. This attribute is standard HTML and should not be stripped. This pull request adds the `slot` attribute to the safe list.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/slot


cure53:
LGTM, thx!

PR #537, opened 2021-05-07
fix: handle invalid xml

NateScarlet:
> This pull request fixes syntax error when input is invalid xml

### Background & Context

DOMPurity throws syntax error if input is empty string and namespace is not html.

Error reason is DOMPurify convert empty string to `<!-->`  which is not valid xml.

image


cure53 (approved):

cure53:
This looks great, thank you!

cure53:
Hmm, this is a strange test result:

https://github.com/cure53/DOMPurify/runs/2525017642?check_suite_focus=true#step:9:280

I can adjust the test suite but it feels wrong, any idea how we can address this? @NateScarlet

NateScarlet:
IE seems not validate on xml.

We can:

a. Avoid `<!-->`, just return empty string if user want string, empty DOM object if user want DOM object.
b. Do validate on our side,

NateScarlet:
test can pass if i just remove the `<!-->` related code

```patch
--- a/src/purify.js
+++ b/src/purify.js
@@ -1104,12 +1104,6 @@ function createDOMPurify(window = getGlobal()) {
     let currentNode;
     let oldNode;
     let returnNode;
-    /* Make sure we have a string to sanitize.
-      DO NOT return early, as this will return the wrong type if
-      the user has requested a DOM object rather than a string */
-    if (!dirty) {
-      dirty = '<!-->';
-    }

     /* Stringify, in case dirty is an object */
     if (typeof dirty !== 'string' && !_isNode(dirty)) {
```

cure53:
No, we cannot do this, some browsers will act out when we do that.

NateScarlet:
Maybe add a flag to indicate empty input and handle as empty string when set xml innerHTML?

```patch
diff --git a/src/purify.js b/src/purify.js
index e83041d..4322af2 100644
--- a/src/purify.js
+++ b/src/purify.js
@@ -324,6 +324,7 @@ function createDOMPurify(window = getGlobal()) {
   const HTML_NAMESPACE = 'http://www.w3.org/1999/xhtml';
   /* Document namespace */
   let NAMESPACE = HTML_NAMESPACE;
+  let IS_EMPTY_INPUT = false;
 
   /* Keep a reference to config to pass to hooks */
   let CONFIG = null;
@@ -667,6 +668,9 @@ function createDOMPurify(window = getGlobal()) {
     /* Create a HTML document */
     let doc;
     let leadingWhitespace;
+    if (IS_EMPTY_INPUT){
+      dirty = '';
+    }
 
     if (FORCE_BODY) {
       dirty = '<remove></remove>' + dirty;
@@ -1109,6 +1113,7 @@ function createDOMPurify(window = getGlobal()) {
       the user has requested a DOM object rather than a string */
     if (!dirty) {
       dirty = '<!-->';
+      IS_EMPTY_INPUT = true;
     }
 
     /* Stringify, in case dirty is an object */
```

cure53:
That looks okay, imho!

cure53:
You wanna spin up a PR? Happy to review and merge. Thanks!

PR #538, opened 2021-05-07
fix: handle empty input

NateScarlet:
> This pull request fixed wrong result when process empty string in IE

### Background & Context

https://github.com/cure53/DOMPurify/pull/537#issuecomment-834144774


cure53:
Fantastic, thank you :)

PR #541, opened 2021-05-26
Fix issues caused by missing config reset

NateScarlet:
> This pull request fix issues caused by missing config reset...

### Background & Context

`NAMESPACE` and `IS_EMPTY_INPUT` incorrectly persisted between runs


cure53:
Amazing, thanks a lot!

Issue #567, opened 2021-09-13
Support strict XHTML documents


tosmolka:
> This issue proposes a feature which adds support for sanitization of strict XHTML documents. 

### Background & Context

I would like to use DOMPurify to sanitize strict XHTML documents. In contrary to HTML, tag and attribute names  in XHTML documents are case-sensitive. This is somehow related to #144 or #476. I believe current version of DOMPurify treats all input as HTML (tag and attribute names are NOT case-sensitive and converted to lowercase). 

Say XHTML document is served with Content-Type application/xhtml+xml` and contains <A> tag. Current browsers (I tested with latest Chrome) will ignore such element because tags must be lowercase in XHTML. If we sanitize such document through DOMPurify before serving it to the browser - <A> will get converted to <a> and browser renders it as a link. I would instead expect DOMPurify to keep only valid (lowercase) tags and therefore remove <A>. 

### Feature

We are proposing an implementation that adds new option - PARSER_MEDIA_TYPE which can be:
 - text/html (default, current behavior)
 - application/xhtml+xml (new behavior, DOMPurify parses input as strict XHTML and tag and attribute names are case-sensitive)

cure53:
Heya, thanks for adding the PR. We had a look at the code and have no concerns, tests are green as well, lgtm :)

cure53:
Alright, linter is happy as well, great job :partying_face: You need a release for this soon?

tosmolka:
Thank you. This can wait on standard release, say couple of weeks is fine. This change might need more testing prior release anyway. 

I also noticed this error (among others) in https://github.com/cure53/DOMPurify/runs/3595585691 , I think TITLE should not be in the output, could be an issue with the implementation. 

```
Expected: "<html xmlns=\"http://www.w3.org/1999/xhtml\"><head></head><body>invalid<a href=\"#\">valid</a></body></html>"
Actual: "<html xmlns=\"http://www.w3.org/1999/xhtml\"><head></head><body>invalid<a href=\"#\" TITLE=\"title\">valid</a></body></html>"

```

cure53:
Yeah, I am working on this right now :) Should be fixed soon.

cure53:
Alright, all fine :D Will release soon.

PR #568, opened 2021-09-13
Add PARSER_MEDIA_TYPE option to support strict XHTML documents

tosmolka:
> This pull request implements feature #567 (Support strict XHTML documents)

### Background & Context

DOMPurify assumes input is HTML and forces text/html media type when the input is parsed with DOMParser. We would like to have the ability to parse the input as XHTML with media type application/xhtml+xml. This will make DOMParser even more strict and for example allow only lowercase tag and attribute names. 

### Tasks

- Added new option PARSER_MEDIA_TYPE
  - text/html (default, current behavior)
  - application/xhtml+xml (new behavior, DOMParser treats input as xhtml and tags/attributes are case sensitive)
- Added test cases "Config-Flag tests: PARSER_MEDIA_TYPE"


cure53:
This looks great, thank you very much! Running tests now...

PR #572, opened 2021-09-16
Fix README.md to reference application/xhtml+xml

zcorpan:
> This pull request fixes README.md for `PARSER_MEDIA_TYPE`

### Background & Context

https://github.com/cure53/DOMPurify/pull/568 introduces XML parsing by specifying application/xhtml+xml as `PARSER_MEDIA_TYPE`. Any other type is parsed as text/html, if I understand correctly. The README.md said to use application/xml for XML parsing, which is thus incorrect.

### Tasks

none

### Dependencies

none


cure53:
Well spotted, thanks!

PR #588, opened 2021-10-19
Optimize `_isClobbered` by only checking `form`s

### LINKED ISSUE #585: Questions About `_isClobbered`
Out of curiosity, could you give an example of some input that makes the condition at https://github.com/cure53/DOMPurify/blob/dbf7a34152b9502ef0b828f194dcdd8bd54282a0/src/purify.js#L793 return `true`?

Also, why does this still run when the `SANITIZE_DOM` config option is `false`? I was under the impression that setting `SANITIZE_DOM` to `false` disables any clobbering-related checks. Is that not the purpose of `SANITIZE_DOM`? If not, what's the purpose?

GrantGryczan:
Resolves #585


cure53:
Aye, are the tests all passing with this PR?

GrantGryczan:
Allow me to check. And I will also fix the merge conflicts. The same goes for my other PR.

cure53:
Thanks :)

GrantGryczan:
I'm realizing there would still be conflicts between the two PRs. Not fun to deal with. I'll just create a new PR with the combined changes.

PR #590, opened 2021-10-19
Only `importNode` if `shadowroot` is allowed

### LINKED ISSUE #584: Question: Why not `{ RETURN_DOM_FRAGMENT: true, RETURN_DOM_IMPORT: false }`?
The readme states there are XSS vulnerabilities when `RETURN_DOM_FRAGMENT` is `true` while `RETURN_DOM_IMPORT` is `false`. Out of curiosity, could you give any example(s) of this? And an explanation of why it occurs or why `RETURN_DOM_IMPORT` prevents it would be appreciated :)

GrantGryczan:
Resolves #584


GrantGryczan:
Ah, no, I totally understand the absolute need for thorough testing in this case! It's very trivial here since this is such a security-sensitive library. I just didn't realize there was something I needed to explicitly do in order to abide by that beforehand. :)

GrantGryczan:
I'm unsure of what specifically I'm supposed to create a test for, as there is already a test here for exactly the case you linked from JS Bin:

https://github.com/cure53/DOMPurify/blob/main/test/test-suite.js#L494-L498

Does that not cover this already?

GrantGryczan:
I need your confirmation on this: I think it would be better to have it always call `importNode` if `RETURN_DOM_IMPORT` is set to `true`, so that it works as expected when set explicitly, but have it _default_ to `false` when `shadowroot` is not in the `ALLOWED_ATTR` attribute, rather than overriding it and preventing `importNode` from getting called when it isn't. Does that sound right?

GrantGryczan:
Here's what I mean:

https://github.com/GrantGryczan/DOMPurify/commit/9fa434cea19451a84cb61c2f9529c057705a924a

PR #597, opened 2021-11-23
Allow passing of ALLOWED_CUSTOM_ELEMENTS config regex or predicate

### LINKED ISSUE #596: DOMPurify removes all custom elements
In the almost-default configuration (except `RETURN_TRUSTED_TYPES: true`) DOMPurify strips out all custom elements, making it unusable for custom element-based applications.

It also removes the attribute values of `is`-attributes, making derived custom elements unusable.

Here's a reproducer repo demonstrating the issue: https://github.com/franktopel/dompurify-custom-elements

### Background & Context

Doing something as simple as `document.body.innerHTML = '<foo-bar></foo-bar>';` with a default policy in place will set `''` as `innerHTML`.

This is my trusted-types-defining script:

```js
// trusted-types-policies.js
import DOMPurify from 'dompurify';
DOMPurify.setConfig({
  RETURN_TRUSTED_TYPE: true,
});
const policies = {
  createHTML: (str) => {
    return DOMPurify.sanitize(str);
  },
  createScriptURL: (str) => str,
  createScript: (str) => str,
};

if (window.trustedTypes && window.trustedTypes.createPolicy) {
  window.trustedTypes.createPolicy('default', policies);
} 
```

The HTML defines this Content-Security-Policy:

```html
<meta http-equiv="Content-Security-Policy"
    content="require-trusted-types-for 'script'; trusted-types dompurify default" />
```

### Bug

DOMPurify removes any custom elements in the sanitized `TrustedHTML`.

#### Input

`<foo-bar></foo-bar><div is="custom-div"></div>`

#### Given output

`<div is=""></div>`

#### Expected output

`<foo-bar></foo-bar><div is="custom-div"></div>`

### Feature

Without this, it's impossible to efficiently work with any sort of view components that import/load their HTML from a HTML file (as a string, as we won't actually have HTML imports for quite some time).

In the current documentation, I wasn't able to find a way to configure DOMPurify to allow any custom elements. Whitelisting those tags won't work for me as the trusted types policy is implemented in the atomic web component library which is being consumed by the application, which itself is made up from composed custom elements loading HTML that includes a lot of the atomic, and even other composed custom elements.

If whitelisting is the only option, at least that would have to allow all tags based on a pattern, like all custom elements whose tag name starts with a certain string (as well as any `is`-attribute values following the same pattern).

franktopel:
> This pull request fixes #596.

### Background & Context

DOMPurify needs to be configurable to allow custom elements to survive sanitization.

It does this now via a new configuration option `ALLOWED_CUSTOM_ELEMENTS` which accepts either a regular expression that will then be used to test the element's `tagName` (or the value of an `is`-attribute in case of derived custom elements), or a predicate function that does the test. The default value for this new option is `null`, which means no change in behaviour (custom elements are removed and is-attribute-values are emptied during sanitization).

#### Example 1:

```js
DOMPurify.setConfig({
  ALLOWED_CUSTOM_ELEMENTS: /^(custom-|foo-)/,
});
```

will allow any custom elements as long as the `tagName` starts with either `custom-` or `foo-`, or if it is a derived built-in, it will be allowed if the value of the `is`-attribute  starts with either `custom-` or `foo-`.

#### Example 2:

```js
DOMPurify.setConfig({
  ALLOWED_CUSTOM_ELEMENTS: (tagName) => tagName.startsWith('custom-') || tagName.startsWith('foo-'),
});
```

achieves the same result as in the regex example above.

### Tasks

- [ ] prettier doesn't like the code formatting so I was forced to commit using `--no-verify`.

### Dependencies

- [x] none


franktopel:
I squashed all the commits into one.

cure53:
Some tests still fail, looks like lint isn't happy?

cure53:
The changes look good, once the linter is happy, I am ready to merge and have another close look.

franktopel:
I mentioned this in the PR:

image

Not sure how to address this.

cure53:
> --fix                          Automatically fix problems

https://eslint.org/docs/user-guide/command-line-interface 

That will do the trick :)

franktopel:
There is no `.eslintrc`.

franktopel:
I was able to make eslint and prettier happy. Will squash commit that.

cure53:
Awesome, thanks! Was just gonna respond about the "--" option

cure53:
Alright, let's see what the tests say now :)

franktopel:
Linter should be happy now, and tests should pass. One warning about code complexity remains, but that was already there before I forked your code.

franktopel:
Btw, `npm run fomat:js` applies some more fixes in karma-related files which I haven't committed.

franktopel:
I can push the proof-of-concept version of the reproducer repo, if you want. Otherwise I'm deleting that.

cure53:
> Btw, npm run format:js applies some more fixes in karma-related files which I haven't committed.

I'll check those, no worries

cure53:
Having a closer look on my system now, thanks a lot so far :)

PR #599, opened 2021-11-24
Significant general performance optimizations in some cases

### LINKED ISSUE #584: Question: Why not `{ RETURN_DOM_FRAGMENT: true, RETURN_DOM_IMPORT: false }`?

### LINKED ISSUE #585: Questions About `_isClobbered`

GrantGryczan:
Resolves #584
Resolves #585 

Incidentally, this also ensures `importNode` is only called on the output when necessary, rather than letting users configure it with `RETURN_DOM_IMPORT`, possibly opening themselves up to vulnerabilities. (See discussion below.)


GrantGryczan (commented):
I need your confirmation on the last commit:

I think it would be better to have it always call `importNode` if `RETURN_DOM_IMPORT` is set to `true`, so that it works as expected when set explicitly, but have it _default_ to `false` when `shadowroot` is not in the `ALLOWED_ATTR` attribute, rather than overriding it and preventing `importNode` from getting called even when `RETURN_DOM_IMPORT` is `true`. Does that sound appropriate?

If so, some tests will need to be changed in consideration of this change, but I can probably do that.

cure53:
Alright, let's run some tests :)

GrantGryczan:
It's not ready yet; see my last comment.

cure53:
Oh, sorry, no worries.

> I think it would be better to have it always call importNode if RETURN_DOM_IMPORT is set to true, so that it works as expected when set explicitly, but have it default to false when shadowroot is not in the ALLOWED_ATTR attribute, rather than overriding it and preventing importNode from getting called even when RETURN_DOM_IMPORT is true. Does that sound appropriate?

So, the goal of whatever we do here is to avoid having people configure themselves to be prone to XSS when they are not aware of it. No matter how we do it, this needs to be prio number one.

GrantGryczan:
> So, the goal of whatever we do here is to avoid having people configure themselves to be prone to XSS when they are not aware of it. No matter how we do it, this needs to be prio number one.

In that case, perhaps it would be better to discard the `RETURN_DOM_IMPORT` option entirely, in favor of forcing it to `importNode` when the `shadowroot` attribute is allowed?

cure53:
Hmmm, not sure how many people already use it. And not sure how many of those allow `shadowroot` :D

GrantGryczan:
Of the people who are using it, there are two possible cases:
1. They allow `shadowroot`. Then removing the option fixes an XSS vulnerability for them.
2. They don't allow `shadowroot`. Then removing the option changes nothing and the config option is a no-op for them.

I don't see any issue with removing it.

cure53:
I think you are right, let's nuke it.

GrantGryczan:
Oh! I realize the reason it's there despite your concern about users unwittingly opening themselves to vulnerability. It's useful if all of the following conditions are true for your use case:

1. You have `RETURN_DOM` or `RETURN_DOM_FRAGMENT` enabled (obviously).
2. You allow the `shadowroot` attribute.
3. You are NOT appending the DOM node outputted by DOMPurify into any document. (This one is the one we forgot about.)
4. You want to improve performance by disabling the `importNode` call.

I still don't think it's worth keeping though, because it's a minor benefit (performance improvements for very large node trees) and only in very rare conditions (`shadowroot` allowed, not appending to any document). But it's up to you; let me know what you think.

cure53:
Let's remove it for now and see if people complain. Less complexity is better in the long run.

GrantGryczan:
Alright, all ready now. I've changed up a lot of the tests which used to pertain to DOM importing. Needless to say, review my changes to the tests especially carefully.

README.md
```diff
@@ -342,7 +332,7 @@ Many people helped and help DOMPurify become what it is and need to be acknowled
...
 
-## Testing powered by
```
GrantGryczan, line 345: Haha, looks like I unwittingly removed a couple instances of trailing whitespace. I have a shortcut configured in my editor to remove trailing whitespace automatically on save.

GrantGryczan:
Looks like there are merge conflicts now that you've accepted a PR from someone else in the meantime. I hope it's easier for you to take care of that.

cure53:
There should be nothing that conflicts with your changes afaics, just generated stuff

cure53:
Alright, conflicts resolved, let's see what the tests say!

cure53:
This looks great, thank you very much :bow: :smile:

PR #600, opened 2021-11-24
added tests for ALLOWED_CUSTOM_ELEMENTS config option

franktopel:
Added tests for the fix for #596.


cure53:
Looking great, thank you :)

PR #602, opened 2021-11-24
allow arbitrary attributes and their values for configured custom elements

### LINKED ISSUE #601: Followup for #596: Preserve all attributes on autonomous custom elements
This is a followup for #596. Background see that issue.

### Bug

Given we configure DOMPurify with `{ ALLOWED_CUSTOM_ELEMENTS: /^foo-/ }`, it must preserve any attributes and attribute values on those matching custom elements.

#### Input

`<foo-bar baz="foobar"></foo-bar>`

#### Given output

`<foo-bar></foo-bar>`

#### Expected output

`<foo-bar baz="foobar"></foo-bar>`

### Feature

Custom elements can implement their own attributes and are free to name them however they like. For these elements to still work, not changing their attributes is essential.

franktopel:
* also added basic custom element tagname check so we don't open the floodgates with `ALLOWED_CUSTOM_ELEMENTS: /script/`

This pull request fixes #601, which is a followup for #596.

### Dependencies

If there are any dependencies on PRs or API work then list them here.

- [x] no dependencies


franktopel:
Did I mention I hate prettier? 🗡️

cure53:
Good morning, just quick FYI, I removed the conflicts, no other PRs expected from my end today, so things should stay conflict-free :)

test/test-suite.js
```diff
@@ -593,15 +593,15 @@ module.exports = function (DOMPurify, window, tests, xssTests) {
     );
     assert.equal(
       DOMPurify.sanitize(
-        '<foo-bar baz="foobar"></foo-bar><div is="foo-baz"></div',
-        { ALLOWED_CUSTOM_ELEMENTS: (tagName) => tagName.startsWith('foo-') }
+        '<foo-bar bas="foobar"></foo-bar><div is="foo-baz"></div>',
```
franktopel, line 596: `bas` should be `baz`

test/test-suite.js
```diff
@@ -593,15 +593,15 @@ module.exports = function (DOMPurify, window, tests, xssTests) {
...
+        { ALLOWED_CUSTOM_ELEMENTS: (tagName) => tagName.match(/^foo-/) }
       ),
       '<foo-bar baz="foobar"></foo-bar><div is="foo-baz"></div>'
     );
     assert.equal(
       DOMPurify.sanitize(
-        '<foo-bar baz="foobar"></foo-bar><div is="foo-baz"></div>',
-        { ALLOWED_CUSTOM_ELEMENTS: (tagName) => tagName.endsWith('-bar') }
+        '<foo-bar bas="foobar"></foo-bar><div is="foo-baz"></div>',
```
franktopel, line 603: `bas` should be `baz`

franktopel:
> Good morning, just quick FYI, I removed the conflicts, no other PRs expected from my end today, so things should stay conflict-free :)

Thank you for that, but some changes you introduced make two tests fail, see my review comments above.

test/test-suite.js
cure53: Should I fix real quick?

franktopel: Yes, please. I'd be happy not to have to do another fork and PR for this.

franktopel:
I still have the conflicted version on my hard drive, not sure how to test without fixing that first.

franktopel:
Just tested that agains our "components onepager", looking fine to me.

cure53:
What risks do you presently see here - i.e. how would the existing check be bypassable?

franktopel:
I'm not a security person, and thus, I have no idea. But I know there's alot of very creative ways to bypass security measures implemented by people whose main concern isn't security.

cure53:
So, what I think will bite us in the ass right now is this: https://w3c.github.io/mathml-core/#dfn-annotation-xml

Because, this would be falsely detected as a custom element, no?

franktopel:
The basic check would let that pass, because it only detects the presence of a `-`. How is that a problem?

cure53:
Not sure yet, but it's _not_ a custom element but it will be detected as one. And Firefox fully supports it via MathML.

Further, are we sure we want to really allow _any_ attribute name for custom elements?

franktopel:
> Further, are we sure we want to really allow _any_ attribute name for custom elements?

Probably we ideally wouldn't, but remember all this only happens if a user decides they need/want that.

cure53:
> > Further, are we sure we want to really allow _any_ attribute name for custom elements?
> 
> Probably we ideally wouldn't, but remember all this only happens if a user decides they need/want that.

Correct, but the user has likely no idea that this exists. Also, will the user be aware that this will work?

```javascript
DOMPurify.sanitize(
    '<annotation-xml><x-y onclick=alert(1)></x-y></annotation-xml>', 
    {ALLOWED_CUSTOM_ELEMENTS: /\w+-\w+/ }
)
```

cure53:
> Also, if the attribute is in the forbid list, isn't it still eradicated?

There is no forbid-list. We only have an allow-list.

franktopel:
It should be quite obvious to everyone that this will open the floodgates: `/\w+-\w+/`

cure53:
I am talking about the attributes, mostly

franktopel:
I'm attending my daily now, will come back at you.

franktopel:
A regex or predicate for that imo would only serve one purpose: It would make it clearer to the user what they are allowing. It is practically impossible to whitelist those attributes.

On the other hand it would also slow down the purification process, and it would make the whole feature harder to use. It's you library, so it's up to you to decide.

I just need a solution that allows me to a) use DOMPurify with custom elements, and through this, allows me b) to enable Trusted Types.

cure53:
So, right now we allow any attribute in case an element passes a certain check. This cannot stay as is. We *must* that it is not possible too use custom elements to smuggle in risky attribute, there is no way around that.

> If that helps, we could check if lcName matches /^on/, but that would also eradicate attributes like once="true", which is likely to produce follow-up issues.

This is a terrible idea and we cannot do that.

cure53:
I see only one way here: Re-enforce the attribute allow-list for custom elements and give a developer the chance to allow-list more (we have that already) and potentially define a wider pattern (i.e. via regex or alike).

franktopel:
If you want to discuss the issue on the phone, feel free to give me a call. Find my contact details on https://connexo.de/kontakt.php

cure53:
Thanks for the good call, proposal currently is:

* Additional config option to specify a regex for attribute names (default: allow-list is enforced)
* Rename existing config option to something more fitting (maybe `CUSTOM_ELEMENT_ALLOW_TAG`, `CUSTOM_ELEMENT_ALLOW_ATTR`)

franktopel:
Bundling related config options in an object would be what I would do:

```js
CUSTOM_ELEMENT_HANDLING: Object.seal({
  tagNameCheck: /^(vaadin-|ionic-)/,
  attributeNameCheck: /w+/,
  allowDerivedElements: true
})
```

cure53:
Hmmm, this looks reasonable.Works for me! As long as the defaults are strict and prohibitive, this should be good.

franktopel:
Just out of curiosity, why is prohibiting `/^on/` a terrible idea?

cure53:
a) Block-lists are by design dysfunctional in security
b) There is way more attributes that execute JavaScript than `on*` (`href`, `action`, `formaction`, values, etc.)

franktopel:
I assume I can't make use of `Object.assign` since it's not supported in any IE?

cure53:
Yes, I kicked MSIE10 out last night but MSIE11 we cannot give up on yet.

franktopel:
Does DOMPurify in certain cases change the order of attributes?

Getting this test fail:

```
not ok 315 - expected: "<foo-bar baz=\"foobar\" forbidden=\"true\"></foo-bar><div is=\"\"></div>", got: "<foo-bar forbidden=\"true\" baz=\"foobar\"></foo-bar><div is=\"\"></div>"
```

cure53:
It is possible that browsers to that in several situations.

franktopel:
Can I rely on that order and adjust `expected` accordingly?

cure53:
I would say so, yes - but sometimes, we sadly need to add several expectations to cover that. Ugly but necessary.

franktopel:
If DOMPurify output would not be deterministic in all cases, I guess we'd have bigger problems.

cure53:
Well, we are a sanitizer, not a markup manager. And sometimes, browsers mess with the order of things, attribute names, etc. etc. - nothing much we can do about that.

franktopel:
How are browsers involved in this? I was assuming DOMPurify processes Strings and outputs Strings.

cure53:
Completely faulty assumption, we utilize the browser to create a document which we sanitize. Then we return the sanitized document as string, node, fragment or whatever people like :)

franktopel:
Interesting, how is that document creation not unsafe in and of itself?

cure53:
https://developer.mozilla.org/en-US/docs/Web/API/DOMParser

franktopel:
Very interesting, since afaik `DOMParser.parse` itself is an XSS sink that is being secured using TrustedTypes.

cure53:
Nope :)

franktopel:
Ah, it's just [Domparser.parseFromString](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString) that is an XSS sink.

cure53:
It's not an XSS sink :D

franktopel:
Well Google says so in that article.

cure53:
We're going off-topic here. The way we use DOMParser is safe and needs no review at the moment.

franktopel:
I've raised an issue on web.dev. https://github.com/GoogleChrome/web.dev/issues/6890

franktopel:
Back to our original topic, tests are all green now and expected results fit. Creating a PR now.

franktopel:
PR is ready and good to go. One build job fails because `yarn` times out. Not sure what to do about that.

PR #603, opened 2021-11-25
switched custom element handling to an object defining three differen…

### LINKED ISSUE #601: Followup for #596: Preserve all attributes on autonomous custom elements

franktopel:
…t properties: tagNameCheck, attributeNameCheck, allowCustomizedBuiltInElements

> This pull request fixes #601 

### Background & Context

We agreed on [this configuration format](https://github.com/cure53/DOMPurify/pull/602#issuecomment-979044494).

### Dependencies

- [x] no dependencies


franktopel:
Not sure what is the problem here. Your repo had a `package-lock.json` before:

image

cure53:
LOL, okay - that happens some times, had that before :)

cure53:
This looks good to me, merging now and running some local tests. Thank you :)

franktopel:
Not sure if my tests are sufficient now, please check.

PR #609, opened 2021-12-13
Revert "Allow SVG filter effects by default"

cure53:
Reverts cure53/DOMPurify#608


cure53:
Additional review needed, we can do this more elegantly me things, @freddyb

PR #611, opened 2021-12-29
chore: updated nvmrc

is2ei:
Updated `.nvmrc`.

### Background & Context

Node.js version 13.x is not supported now.
I updated it to `lts/*` so that we don't have to care about updating it in the future.


cure53:
Excellent, thank you!

PR #614, opened 2022-01-01
chore: updated comments to use main instead of master

is2ei:
updated comments to use `main` instead of `master`

### Background & Context

This project's default branch is `main`.


cure53:
Thanks!

Issue #621, opened 2022-01-19
node.outerHTML = emptyHTML causes TT Sink violation


tosmolka:
### Background & Context

We are using DOMPurify with enforced Trusted Types (`require-trusted-types-for 'script'`) and we started seeing TT Sink violations from this location:

https://github.com/cure53/DOMPurify/blob/262a050fb353911b79977404e93831a11cbd528a/src/purify.js#L698

#### Input

Consider this demo HTML file:

```html
<!doctype html>
<html>
    <head>
        <meta http-equiv="Content-Security-Policy" content="require-trusted-types-for 'script';">
        <script src="../dist/purify.js"></script>
    </head>
    <body>
        <div id="sanitized-dom"></div>
        <script>
            var dirty = '<p onclick=alert(2)>HELLO<iframe srcdoc="<script>alert(1)</scr'+'ipt>"></ifrAMe><br>goodbye</p>';

            DOMPurify.addHook('uponSanitizeElement', function(node, data) {
                if (node.tagName === 'IFRAME') {
                    node.parentNode.removeChild(node);
                }
            });

            document.getElementById('sanitized-dom').appendChild(DOMPurify.sanitize(dirty, {RETURN_DOM: true}));
        </script>
    </body>
</html>
```

#### Given output

DOMPurify returns valid output but line `node.outerHTML = emptyHTML` throws Trusted Types sink violation `This document requires 'TrustedHTML' assignment.`. 

This is because DOMPurify is trying to force-remove (method `_forceRemove`) a node that has no parent (it was already removed in custom hook) and tries to assign empty string (not empty TrustedHTML) into outerHTML property. 

This is a very real example from our application, it took us a while to debug and find the root cause. 

#### Expected output

DOMPurify returns valid output without causing any Trusted Types sink violations. I think the fix here is to always use emptyHTML from TT policy if such policy is available.

cure53:
Interesting find, thanks for flagging and fixing! The PR looks good, checking tests appear to be happy as well.

PR #622, opened 2022-01-19
fix: node.outerHTML = emptyHTML causes TT Sink violation

tosmolka:
> This pull request fixes corner-case with Trusted Types when `node.outerHTML = emptyHTML` causes TT Sink violation

### Background & Context

This is a fix proposal for https://github.com/cure53/DOMPurify/issues/621 . This changes `emptyHTML` to be always constructed from TT policy if such policy is available.


tosmolka:
Can you also pls double-check generated files? I would prefer not committing into `dist/` but pre-commit hook forced me.

cure53:
This looks good, thanks a lot!

PR #628, opened 2022-01-23
chore: remove deprecated package babel

is2ei:
This PR removes deprecated package `babel`.

### Background & Context

The package has been deprecated.

https://www.npmjs.com/package/babel
> In 6.x, the babel package has been deprecated in favor of babel-cli. Check https://opencollective.com/babel to support the Babel maintainers

Also, it seems this project doesn't depend on the package so I guess we can remove it simply.


cure53:
Thank you!

PR #630, opened 2022-01-24
chore: bump karma-chrome-launcher

is2ei:
This PR updates `karma-chrome-launcher` version to the latest version.

### Background & Context

https://github.com/karma-runner/karma-chrome-launcher/blob/master/CHANGELOG.md


is2ei:
Yes. It seems there are no breaking changes though the major version is incremented.

tdeekens:
> Yes. It seems there are no breaking changes though the major version is incremented.

That's how we like semantic versioning 😋.

PR #647, opened 2022-02-13
chore: bump qunit

is2ei:
This PR updates the outdated package `qunit`.

### Background & Context

- `qunit-parameterize` is not compatible with QUnit v2 so it should be removed. Ref: https://github.com/AStepaniuk/qunit-parameterize/issues/15
- QUnit v2 provides an upgrade guide: https://qunitjs.com/upgrade-guide-2.x/

### Tasks

- Tests should pass.


is2ei:
I'd check the CI failure.

is2ei:
Ok. The CI passed. Could you review it?

tdeekens:
This is great! Just for me to follow along: how do we make up for removing the parametrize plugin?

is2ei:
@tdeekens 
Thanks for your comment.
The parametrize plugin provides the ability to specify the list of test cases.
I found the plugin is no longer maintained.
According to [the README](https://github.com/AStepaniuk/qunit-parameterize/blob/master/README.md), the plugin does 3 things:
1. parameter assertion
2. title suffix addition
3. skipping test case

I guess we don't need No.3 since there are test cases to skip.
We can use string interpolation for No.2.
For No.1, I think `forEach` can be the replacement.

cure53:
Epic, thanks! :D

PR #659, opened 2022-03-06
Stop trimming value attributes

christopherehlen:
> This pull request changes how value attributes are trimmed in the input element

### Background & Context

I'm using nano morph to merge DOM changes after performing a DOMPurify.  The value attribute, i.e., user input, is trimmed indiscriminately.  So DOMPurify is changing the user's intended content.

### Tasks

- Stop trimming the value attribute.

### Dependencies

If there are any dependencies on PRs or API work, list them here.

- [x] Resolved dependency
- [ ] Open dependency


cure53:
This looks good, thank you!

PR #666, opened 2022-03-23
chore: fix link to AMD

is2ei:
This PR updates the link to AMD.

### Background & Context

https://github.com/amdjs/amdjs-api/wiki/AMD
> This is a copy of the ["AMD" document in the repo](https://github.com/amdjs/amdjs-api/blob/master/AMD.md), kept here to maintain historical links. If this document differs from the one in the repo, the repo version is the correct one.


cure53:
Thank you!

PR #675, opened 2022-04-24
chore: use import syntax in rollup config

is2ei:
This PR changes `rollup.config.js` to use `import` syntax.

### Background & Context

The official document uses `import` syntax.

e.g. https://rollupjs.org/guide/en/#using-output-plugins


is2ei:
Sorry, I totally forgot about the Karma config file...

PR #680, opened 2022-05-11
Minor performance improvements

tiny-ben-tran:
## Summary
Minor performance improvement for createDOMPurify()

## Background & Context

The engineers at Tiny made some improvements to the DOMpurify and would like to make a contribution upstream so we don't have to maintain these.


cure53:
This looks good, thank you! :smile:

PR #681, opened 2022-05-11
Remove attribute only when necessary

tiny-ben-tran:
## Summary
Remove an attribute if it falls into one of the following cases:
   - not approved by `uponSanitizeAttribute` hook
   - the attribute is invalid
   - it is an invalid data-* attribute
   
## Background & Context

The engineers at Tiny made some improvements to the DOMpurify and would like to make a contribution upstream so we don't have to maintain these.


cure53:
That one seems problematic I think, see test results cc @tiny-ben-tran

PR #692, opened 2022-06-06
Make TAG and ATTR cfg options case-sensitive when parsing XHTML

tosmolka:
## Summary
A while back we introduced PARSER_MEDIA_TYPE option for parsing XHTML files where tag and attribute names are case-sensitive. 

We have one use case where we would like DOMPurify to keep certain custom tags and attributes by specifying ALLOWED_TAGS and ALLOWED_ATTR options. The problem is that `addToSet` function forces lowercase by calling `stringToLowerCase` on all elements. DOMPurify therefore does not respect ALLOWED_TAGS and ALLOWED_ATTR options and always removes our custom tags and attributes. 

This PR fixes this issue by passing `transformCaseFunc` to `addToSet`, happy to redo if there is a better way to do it.

### DIFF
--- a/src/purify.js
+++ b/src/purify.js
@@ -380,29 +380,55 @@ function createDOMPurify(window = getGlobal()) {
     /* Shield configuration object from prototype pollution */
     cfg = clone(cfg);
 
+    PARSER_MEDIA_TYPE =
+      // eslint-disable-next-line unicorn/prefer-includes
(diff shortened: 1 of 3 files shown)


cure53:
This looks great, thank you!

tosmolka:
@cure53 , do you have a release date for this change? Thanks.

cure53:
Why not now :)

Issue #660, opened 2022-03-07
DOMPurify fails on sanitization of Trusted Types sink attributes


tosmolka:
> Even if attribute is allowed, DOMPurify performs basic sanitization, such as stringTrim. This can cause Trusted Types Sink violation for attributes that are expecting trusted types (e.g. <script src>). 

### Background & Context

DOMPurify does not (by design) validate script URLs and we want to perform our own validation using hooks. It turns out DOMPurify calls setAttribute for all allowed attributes and this causes Trusted Types sink violation:

https://github.com/cure53/DOMPurify/blob/5c28248a9d221d4515527412d68459c6bb4937e9/src/purify.js#L1196

#### Input

```html
<!doctype html>
<html>
    <head>
        <meta http-equiv="Content-Security-Policy" content="trusted-types dompurify; require-trusted-types-for 'script';">
        <script src="https://cdn.jsdelivr.net/npm/dompurify@2.3.6/dist/purify.js"></script>
    </head>
    <body>
        <script type="text/javascript">
          window.addEventListener('load', () => {
            var dirty = `
              <h3>Gallery</h3>
              <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/glightbox/dist/css/glightbox.min.css">
              ${'\x3C'}script src="https://cdn.jsdelivr.net/npm/glightbox/dist/js/glightbox.min.js">${'\x3C'}/script>
              
              <a href="https://biati-digital.github.io/glightbox/demo/img/large/gm2.jpg" class="glightbox" data-title="Test Title">
                <img src="https://biati-digital.github.io/glightbox/demo/img/small/gm2.jpg" alt="image">
              </a>
            `;
            DOMPurify.addHook('uponSanitizeElement', function(node, data) {
              const policy = {
                script: ['https://cdn.jsdelivr.net/npm/glightbox/dist/js/glightbox.min.js'],
                style: ['https://cdn.jsdelivr.net/npm/glightbox/dist/css/glightbox.min.css'],
              };
              if (
                data.tagName === 'script' && !policy.script.includes(node.getAttribute('src')) ||
                data.tagName === 'link' && !policy.style.includes(node.getAttribute('href'))
              ) {
                return node.parentNode.removeChild(node);
              }
            });
            var clean = DOMPurify.sanitize(dirty, {ADD_TAGS: ['link','script'], RETURN_TRUSTED_TYPE: true});
            var iframe = document.createElement('iframe');
[...]

PR #699, opened 2022-07-18
Support sanitization of attributes that require Trusted Types

tosmolka:
## Summary

This PR should fix [660](https://github.com/cure53/DOMPurify/issues/660) for attributes without namespace. 

## Background & Context

We want to use DOMPurify to allow `script.src` attribute and provide custom src validation via hooks. This is not possible today as DOMPurify will call setAttribute with raw string and cause Trusted Types Sink violation. 

This PR leverages [TrustedTypePolicyFactory.getAttributeType()](https://w3c.github.io/webappsec-trusted-types/dist/spec/#dom-trustedtypepolicyfactory-getattributetype) to handle such cases.

### DIFF
--- a/src/purify.js
+++ b/src/purify.js
@@ -55,6 +55,9 @@ const _createTrustedTypesPolicy = function (trustedTypes, document) {
       createHTML(html) {
         return html;
       },
+      createScriptURL(scriptUrl) {
+        return scriptUrl;
+      },
     });
   } catch (_) {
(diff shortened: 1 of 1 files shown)


tosmolka:
Yes please but we can wait a week or two if needed.

cure53:
Which is exactly when I will be on vacation :grimacing: That is why I am asking :smile:

Issue #556, opened 2021-08-07
Add support for removing elements but keeping their children


0xedward:
### Background & Context

Hi! @mozfreddyb and I are working on a polyfill for the [HTML Sanitizer API](https://wicg.github.io/sanitizer-api/). 

Currently, the polyfill shims DOMPurify by taking a [HTML Sanitizer Configuration object](https://wicg.github.io/sanitizer-api/#configuration-object) and converting it to a [configuration DOMPurify accepts](https://github.com/cure53/DOMPurify/blob/HEAD/dist/purify.js#L342-L437), then any `sanitize` calls to the polyfill are pass along to DOMPurify.

### Feature

While working on converting a HTML Sanitizer configuration to DOMPurify configuration, we haven't found a way to support [`blockElements`](https://wicg.github.io/sanitizer-api/#dom-sanitizerconfig-blockelements). We thought that maybe using `FORBID_TAGS` with `KEEP_CONTENT` would get us the functionality we are looking for, but the limitation with that approach would mean we can't set `FORBID_TAGS` without `KEEP_CONTENT` at the same time. So our HTML Sanitizer configuration would either support `blockElements` or [`dropElements`](https://wicg.github.io/sanitizer-api/#dom-sanitizerconfig-dropelements), but not both at the same time.

Is there a configuration option we missed that we could use to achieve desired functionality of `blockElements`? Thanks! :)

cure53:
Heya :) Indeed, I don't think we can do that via config options. But we can do that via hook, would hat be viable for the polyfill?

0xedward:
Thanks for the quick response! :) I'll look into hooks and give it a go. The `*SanitizeElements` hooks looks promising

If we went down the route of adding something like `BLOCK_ELEMENTS` as a configuration option, would that PR be something that the team would be open to reviewing/accepting?

cure53:
Exactly, this one should be the one you need as it fires right before the `KEEP_CONTENT` checks happen. You could check the element and the config and then decide whether to remove child elements based on your config or let the library decide.

cure53:
I had a closer look @0xedward - it seems that this is indeed tough to do cleanly with a hook. I think it might be better to expose the `FORBID_CONTENTS` array to be extendable. 

Assuming we did that, would the below match what you expect in terms of sanitation result?

```javascript
var clean = DOMPurify.sanitize(`
    <div><b>preserve me</b></div>
    <p><b>no not preserve me</b></p>
[...]

PR #709, opened 2022-08-10
Add documentation for `FORBID_CONTENTS`

0xedward:
## Summary

We added support for [removing elements but keeping their children](https://github.com/cure53/DOMPurify/issues/556) in https://github.com/cure53/DOMPurify/pull/557. I had added an example to the README to document the new config option back when I opened https://github.com/cure53/DOMPurify/pull/557, but forgot to make a PR with those changes 😅

## Dependencies

- [x] Resolved dependency
- [ ] Open dependency


cure53:
Awesome, thanks! :)

PR #732, opened 2022-11-14
chore: fix allowCustomizedBuiltInElements value in readme

Pomierski:
## Summary

The value in readme seems to be incorrect. The comment above informs us that if we pass false as a value, then no customized built-ins are allowed.

## Tasks

- xxxx
- xxxx
- xxxx


cure53:
Oh, nice catch, thanks :D

PR #748, opened 2022-12-19
Fix Trusted Types Sink violation with empty input and NAMESPACE

### LINKED ISSUE #747: Trusted Types Sink violation with empty input and NAMESPACE
### Bug

DOMPurify produces unnecessary Trusted Types Sink violation when called with empty input and NAMESPACE option. 

#### Input

```html
<!DOCTYPE html>
<html>
    <head>
        <meta http-equiv="Content-Security-Policy" content="require-trusted-types-for 'script'">
        <script src="https://cdn.jsdelivr.net/npm/dompurify/dist/purify.js"></script>
    </head>
    <body>
        <script>
          DOMPurify.sanitize("", {NAMESPACE: "http://www.w3.org/2000/svg"});
        </script>
    </body>
</html>
```

#### Given output

Following line produces Trusted Types Sink violation as it tries to assign raw string into Element.innerHTML sink:

https://github.com/cure53/DOMPurify/blob/4945074eb6cb4994eedc4d57dd4582f438f8dcf4/src/purify.js#L847

#### Expected output

DOMPurify uses TrustedHTML instead of raw string and no Trusted Types Sink violation is produced

tosmolka:
Use TrustedHTML instead of raw string when assigning empty string to Element.innerHTML sink. This should fix #747 .

### DIFF
--- a/src/purify.js
+++ b/src/purify.js
@@ -844,7 +844,9 @@ function createDOMPurify(window = getGlobal()) {
     if (!doc || !doc.documentElement) {
       doc = implementation.createDocument(NAMESPACE, 'template', null);
       try {
-        doc.documentElement.innerHTML = IS_EMPTY_INPUT ? emptyHTML : dirtyPayload;
+        doc.documentElement.innerHTML = IS_EMPTY_INPUT
+          ? emptyHTML
+          : dirtyPayload;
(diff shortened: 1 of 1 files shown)


cure53:
This looks great, thank you!

PR #765, opened 2023-02-04
Restore config for removing self closing tags in attributes

### LINKED ISSUE #761: [Bug] HTML encoded self closing tags in attribute values leads to the removal of the attribute
When I use html encoded content in an attribute self closing tags leads to the removal of the whole attribute.

### Background & Context

Self closing tags like `<br />` or `` are usually converted to single tags (`<br>`,``), which is totally fine.

#### Input

`<p data-qtip="&lt;br/&gt;">foo</p>`

#### Given output

`<p>foo</p>`

#### Expected output

`<p data-qtip="<br>">foo</p>`

### Working example with single tag

When I'm using single tags, the output is as expected:

#### Input

`<p data-qtip="&lt;br&gt;">foo</p>`

#### Given output

`<p data-qtip="<br>">foo</p>`

edg2s:
Fixes #761


cure53:
This has conflicts, please resolve before we can merge @edg2s

cure53:
It's not clear to me what branch this change should land in. Is MSIE support still needed? If so, we need this in 2.x and 3.x (or main).

AndreVirtimo:
For me MSIE support is not needed. This browser is EOL since last year.

cure53:
Same here, we want to get rid of it asap and more on with 3.x :smile: But not sure about Wikipedia @edg2s

edg2s:
We haven't dropped support for IE11 just yet, so would prefer to do this is in 2.x.

cure53:
Closing this for now in the hopes that the recent changes to 2.x do the trick :smiley: Please reopen in case I overlooked something.

Issue #785, opened 2023-03-27
Possibly drop support for a few more old browsers


buzinas:
### Background & Context

I was upgrading DOMPurify from v2.x to v3.x and when I checked the release notes to check the breaking changes, I realized I shouldn't worry because the only theoretical change was the IE support overhead.

Happy with the outcome, I decided to check how much code was removed, and then for my surprise, Bundlephobia said that this package is 0.3kb bigger now (after minifying / gzipping) than it was before 🤔.

Then I decided to check the code itself, and I realized that by changing `for…in` to `for…of`, a lot of code was added by Babel:
 
https://github.com/cure53/DOMPurify/compare/2.4.5...3.0.1#diff-48108d9bf161db3408a0fc8e8f7ab693247d39db9998ff4cd0d9c3de233dd09aR54-R310

<img width="2513" alt="image" src="https://user-images.githubusercontent.com/7298695/228052685-1473a1c9-f7d8-4ea3-90c9-2b2743a7131c.png">

I think this happens because DOMPurify still supports very old versions of some browsers, eg Chrome 22 (`for…of` was implemented only in Chrome 38, for example).

Is there any reason why these very old browsers are still supported? Especially since we're talking about evergreen browsers.

cure53:
Heya, thanks for filing this. I think indeed in 3.x we can drop support for evergreen browsers, will review your PS next - thanks for that as well :smile:

buzinas:
@cure53 Yes, this was fixed by #786. Thanks!

buzinas:
@cure53 if you can ping me here when a new release is published, please let me know so I can upgrade! 😁

cure53:
ping :)

PR #786, opened 2023-03-27
Bump browsers versions to support for…of and let/const

buzinas:
## Summary

Potential fix for #785.

## Background & Context

Bump browsers versions in order to remove a lot of polyfills for `for…of` and `const/let`.


cure53:
Something seems to break the tests, not sure if related.

https://github.com/cure53/DOMPurify/actions/runs/4536381460/jobs/8000156773#step:7:31

buzinas:
I'll take a look, thanks!

buzinas:
@cure53 the reason tests were failing was due to all target browsers now supporting ESM and then Babel keeping the ES Modules untouched. By removing the `commonjs` plugin, `expect.js` started failing since it was using `module.exports`.

I tried to update it to `export default` instead, which fixed browser tests with Karma, but this change broke JSDOM tests, since they run on NodeJS. Then I converted it to UMD (similar to test-suite.js) and started using `window.tests` on ESM, and although it worked, this felt too hacky.

So I finally renamed `expect.js` to `expect.mjs` and dynamically imported it on `jsdom-node.js`. In order to be able to do that, I had to wrap the code into an async function (since non-modules can't have top-level `await`).

Now all tests work as expected! Please let me know what you think.

cure53:
Fantastic, thank you :smile:

PR #790, opened 2023-04-13
fix: ensure ALLOWED_URI_REGEXP is reset

mukilane:
## Summary

The value of `ALLOWED_URI_REGEXP` is persisted between santize calls without being properly reset.
This pull request fixes it by resetting it `IS_ALLOWED_URI` to its default value instead of its previous value if `ALLOWED_URI_REGEXP` is not sent

## Background & Context

At 490c2efc3b40ed94d02a4758fd0bd6bf8a0dbc24, the `IS_ALLOWED_URI` was set without properly resetting it for subsequent sanitize calls.


cure53:
Oh, nice find, thanks!

mukilane:
Thanks for the quick merge, @cure53 !

Do you happen to have a release date for this change ?

cure53:
Soon. Once I am back from vaycay :smile_cat:

cure53:
Release is out

Issue #573, opened 2021-09-17
Why is the SVG fedropshadow element disabled by default


yisibl:
Is there a way to manually enable disabled elements?

https://github.com/cure53/DOMPurify/blob/bc4f2ec76f288524c2eb9ce27f7cef998fc9ec71/src/tags.js#L196-L205

cure53:
If I remember correctly there was some mXSS risk connected to those. And yes, you can of course enable them - this should work:

```javascript
DOMPurify.sanitize(
    '<svg><feDropShadow>123</feDropShadow></svg>', 
    {ADD_TAGS: ['feDropShadow']}
);
```

yisibl:
I hope some instructions can be added to the document.

cure53:
https://github.com/cure53/DOMPurify#can-i-configure-dompurify

^ all there :)

SelfMadeSystem:
I've searched quite a lot today to try and find this mXSS that uses `feDropShadow.` I can't find one anywhere, not even the mention of one. Do you know anything about it and how it can be sensibly prevented without banning `feDropShadow` elements?

cure53:
At the moment, we also have no knowledge about `feDropShadow` being risky, wanna spin up a PR and add it to the allow-lost?

SelfMadeSystem:
Yeah I could do that. I also have some more questions, but I'll make a new longer issue about it when I Have time.

PR #795, opened 2023-04-26
Moved feDropShadow to the svg filter allowlist

SelfMadeSystem:
## Summary

I moved the `feDropShadow` tag to the allowlist.

## Background & Context

`feDropShadow` is used quite a bit. It was originally but there apparently because of an mXSS risk (https://github.com/cure53/DOMPurify/issues/573#issuecomment-921594876), but after asking about it, he has no knowledge of one (https://github.com/cure53/DOMPurify/issues/573#issuecomment-1522902754)

## Tasks

- [ ] Make sure there really isn't any mXSS


cure53:
LGTM, thanks! :heavy_check_mark:

PR #800, opened 2023-05-01
support TRUSTED_TYPES_POLICY configuration option

dejang:
## Summary

Introducing a new configuration option to allow more flexibility for DOMPurify when used in larger applications which need to handle multiple DOMPurify instances or when allowing a 'dompurify' policy, which DOMPurify attempts to create internally on startup, is not an option.

## Background & Context

The feature has been described in https://github.com/cure53/DOMPurify/issues/798.
The existing behavior has been maintained, DOMPurify will still try to create its internal policy. 

## Tasks
- validate internal 'dompurify' policy is created
- create a custom TrustedTypes policy. Call `DOMPurify.sanitize('', {TRUSTED_TYPES_POLICY: yourPolicy})` and validate call count for the policy method `createHTML` is 2 (one for `emptyHTML` and one when `initDocument` is called)
-  create a custom TrustedTypes policy. Call `DOMPurify.sanitize('', {TRUSTED_TYPES_POLICY: yourPolicy, RETURN_TRUSTED_TYPE: true })` and validate call count for the policy method `createHTML` is 3 (one for `emptyHTML`, one for `initDocument`, one for return value)
- create a custom TrustedTypes policy. Create a new instance of DOMPurify and use `.setConfig({TRUSTED_TYPES_POLICY: yourPolicy})` to configure the policy only. Call `sanitize('')` on the new instance and validate call count  for the policy method `createHTML` is 2 (one for `emptyHTML` and one when `initDocument` is called)
- create a custom TrustedTypes policy. Create a new instance of DOMPurify and use `.setConfig({TRUSTED_TYPES_POLICY: yourPolicy, RETURN_TRUSTED_TYPE: true })` to configure the policy and return type to be a TrustedType. Call `sanitize('')` and validate call count  for the policy method `createHTML` is 3 (one for `emptyHTML`, one for `initDocument`, one for return value)

Issue #798, opened 2023-04-27
Support configurable TrustedTypes policy


dejang:
### Background & Context

We use DOMPurify extensively in a multi tenant web application that serves millions of users. We have also began moving towards TrustedTypes but we find ourselves limited by how the TrustedTypes policy is used within DOMPurify. For starters, we never include DOMPurify via a <script> tag, we always rollup the dependencies in a bundle that is shipped to the browser and is loaded as a <script> tag. Our code is ESM. Additionally, because performance is an influencing factor in our architecture, we use multiple preconfigured instances of DOMPurify in order to skip configuration logic with repeated invocations. The multiple DOMPurify instances have different usages and we'd like to maintain that flexibility in our system. We find that with our current CSP configurations which does not allow duplicates we cannot use more than 1 instance of DOMPurify for the entire application. Given the size of the application and the scenario we are in this affects our ability to maintain our flexibility and, given that with each invocation we need to reset the configuration and the hooks, we end up spending compute cycles that we feel could be avoided.

We would like to request the ability for DOMPurify to support passing in our own TT policy via configuration when instantiated or when the configuration is being set.

### Feature

Given a configuration object passed to `sanitize` or to `setConfig` allow an entry to specify the TT policy that DOMPurify should use.

I've worked on the DOMPurify source code before and would be happy to contribute if this feature is being accepted.

The new configuration option could be named `trustedTypesPolicy` and it must define a `createHTML` hook. The code must validate the object's shape to prevent passing in policies which do not define the minimum requirement.

cure53:
Thank you, that sounds like a very interesting and relevant feature and we'd be happy to review a PR implementing it.

dejang:
Great, I'll get a PR ready for you to review.

cure53:
The PR looks great, thanks, tests are all green as far as I can see. Ready for a new 3.x release? :blush:

tosmolka:
@dejang , I know this is already merged but if I do read the PR correctly then DOMPurify will now always attempt to create default TT policy (dompurify + suffix) even if the new `TRUSTED_TYPES_POLICY` option is set. 

Would you mind minor refactoring to reverse the order and create default policy only if option is not set (or maybe if its invalid)? 

Again, if I read it right, current behavior will cause unnecessary CSP violations when DOMPurify fails to create duplicate default policy on every bundle load. Thanks.

dejang:
[...]

PR #801, opened 2023-05-03
create internal trustedTypes policy only if not specified via config object

dejang:
## Summary

Before https://github.com/cure53/DOMPurify/pull/800:
- DOMPurify would attempt to create an internal Trusted Types policy when loaded on the page.
- No possibility to configure the policy used by DOMPurify.

After https://github.com/cure53/DOMPurify/pull/800:
- DOMPurify would attempt to create an internal Trusted Types policy when loaded on the page.
- `sanitize` and `setConfig` allow switching from the internal policy to a user provided one when `TRUSTED_TYPES_POLICY` configuration option is passed via the configuration object.

This PR:
- DOMPurify will not attempt to create the internal Trusted Types policy when first loaded on the page.
- The internal policy will be created only the first time when calling `sanitize` or `setConfig` *if and only if* `TRUSTED_TYPES_POLICY` configuration option is not specified in the configuration object. This maintains behavior prior to  https://github.com/cure53/DOMPurify/pull/800.
- no change to the behavior introduced in https://github.com/cure53/DOMPurify/pull/800.

## Background & Context

See https://github.com/cure53/DOMPurify/issues/798 for the background discussion.

Note: since the original `loadDOMPurify` test helper was performing a call to `sanitize` without any configuration option this triggered the creation of the internal policy and I could not write a test to validate that there is no policy created when supplying `TRUSTED_TYPES_POLICY` configuration option. I had to refactor the helper to allow writing a new test which specifically looks for the number of policies created when DOMPurify is loaded on the page and after calls to `sanitize`. 

## Tasks

- internal Trusted Types policy creation is skipped if configuration object supplies a Trusted Types policy


src/purify.js
```diff
@@ -130,11 +128,8 @@ function createDOMPurify(window = getGlobal()) {
     }
   }
 
-  let trustedTypesPolicy = _createTrustedTypesPolicy(
-    trustedTypes,
-    originalDocument
-  );
-  let emptyHTML = trustedTypesPolicy ? trustedTypesPolicy.createHTML('') : '';
+  let trustedTypesPolicy;
```
tosmolka, line 131: nit: I would move `let trustedTypesPolicy` declaration just before `if (cfg.TRUSTED_TYPES_POLICY) {` line

src/purify.js
```diff
@@ -130,11 +128,8 @@ function createDOMPurify(window = getGlobal()) {
...
+  let emptyHTML = '';
```
tosmolka, line 132: nit: I would revert to previous declaration 
```
const emptyHTML = trustedTypesPolicy ? trustedTypesPolicy.createHTML('') : '';
```
and move it just after either default or custom trustedTypesPolicy is initialized.

tosmolka (approved):

src/purify.js
dejang: that wouldn't be possible in the present form of the code, the `if` condition is part of the `_parseConfig` function and `let trustedTypesPolicy` is in the upper scope. Moving it before the `if` would make it a local variable to `_parseConfig` and inaccessible to the rest of the methods that look for it on the upper scope. This would be a move involved effort.

src/purify.js
dejang: I believe we would be missing some functionality if we make it a const and initialized one time. Particularly this kind of usage:

```
DOMPurify.sanitize(dirty1);
DOMPurify.sanitize(dirty2, {TRUSTED_TYPES_POLICY: policy});
```

Since the first initialization is without the cfg, which may or may not succeed to produce a policy and at which point `emptyHTML` may end up not being signed, the 2nd call may not work as expected because some methods do rely on `emptyHTML` when they build the DOM tree.

There is, however, an alternative to initializing this. Looking at the existing `trustedTypes` API in Chrome, there seems to be a `trustedTypes.emptyHTML` signed value built into the API so this declaration would not be needed anymore. We could simply declare it as

```
const emptyHTML = trustedTypes? trustedTypes.emptyHTML : ''
```

I did not want to make that change because I do not have enough information to know if the API is considered stable at the moment and if this will be supported. Should be an easy improvement for a follow up PR in case it is.

src/purify.js
tosmolka: Let's keep current change then if it's intentional.

src/purify.js
tosmolka: Let's keep current change then if it's intentional.

cure53:
Cool, are we ready for a merge? :blush:

dejang:
@cure53 let's do it! I'm ready to upgrade whenever the new release is published.

PR #845, opened 2023-08-04
fix(purify): Fix _isNode Function

ssi02014:
## Summary

Hello 👋, @cure53 

I determined that `_isNode` is a function that determines whether the given `object` argument is a `DOM Node`.

Can `typeof Node` result in `object`?
In my tests, `Node` gets "function" as the result of the "typeof" operator.

```js
console.log(typeof Node); // 'function'
```

I checked on `safari`, `firefox`, `Edge`, and `Chrome` and found that the result of typeof Node is `function`.  (Of course, in Node.js, the `undefined`)

https://github.com/cure53/DOMPurify/commit/e7448403cd2602b0448b39a8715f52aba443153b
Looking at the previous commit history, I think it's because of the code like `typeof Node === 'object'` that doesn't work properly.

We simply need to make sure that the result of "typeof Node" is 'function' and that the object is an instance of Node. 

```js
typeof Node === 'function' && object instanceof Node;
```

After fixing the code, the tests all passed. 
Please review this pull request to see if it's appropriate. 🙏

- Firefox
스크린샷 2023-08-05 오전 2 44 36

- chrome
스크린샷 2023-08-05 오전 2 45 28

- safari
스크린샷 2023-08-05 오전 2 46 10

- Edge
KakaoTalk_Photo_2023-08-05-04-15-49

## Tasks

- fix _isNode


cure53:
Thanks again!

I believe the commits back then were mostly needed for MSIE. Since we are phasing out support for MSIE and only support it in the low-frequently maintained 2.x branch, this change for main and 3.x _should_ be fine :sweat_smile:

ssi02014:
@cure53  Thank you!

PR #848, opened 2023-08-09
fix(purify): Remove unnecessary conditional expressions

ssi02014:
## Summary

Hello 👋, @cure53 

Node property "content" is not currently available.
There's no need to conditionally check it, as it always returns `undefined`.

```js
currentNode.content; // Always undefined
currentNode.content.firstElementChild; // Always undefined
```
```js
(!_isNode(currentNode.content) || !_isNode(currentNode.content.firstElementChild)) // Always true
```

I went through all the `NodeType` listed in the documentation below. 
https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType

After fixing the code, the tests all passed, and I think it's safe to remove it from `3.x`.

## Tasks
Remove unnecessary conditional expressions


cure53:
This should be safe, let me do some checks and merge, thanks :slightly_smiling_face:

PR #850, opened 2023-08-10
fix(purify): fix _createIterator

ssi02014:
## Summary

Hello 👋, @cure53 
I was looking at the `_createIterator` function and realized it had an unnecessary argument (`false`), so I did a fix-and-refactor.

https://developer.mozilla.org/en-US/docs/Web/API/Document/createNodeIterator
https://dom.spec.whatwg.org/#dom-document-createnodeiterator

Even if you refer to the documentation above, the `createNodeIterator` function only requires 3 arguments: `root`, `whatToShow`, and `filter`.

스크린샷 2023-08-11 오전 3 52 40
https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L7160

## Tasks

- fix _createIterator


src/purify.js
```diff
@@ -900,19 +896,19 @@ function createDOMPurify(window = getGlobal()) {
...
    *
-   * @param  {Document} root document/fragment to create iterator for
-   * @return {Iterator} iterator instance
+   * @param  {Node} root The root element or node to start traversing on.
+   * @return {NodeIterator} The created NodeIterator
    */
-  const _createIterator = function (root) {
-    return createNodeIterator.call(
-      root.ownerDocument || root,
+  const _createNodeIterator = function (root) {
```
ssi02014, line 904: https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L7160
I've modified the jsdoc with reference to that documentation.

Also, the naming seems clearer for createNodeIterator than createIterator.

src/purify.js
```diff
@@ -900,19 +900,18 @@ function createDOMPurify(window = getGlobal()) {
...
-   * _createIterator
+   * Creates a NodeIterator object that you can use to traverse filtered lists of nodes or elements in a document.
    *
-   * @param  {Document} root document/fragment to create iterator for
-   * @return {Iterator} iterator instance
+   * @param  {Node} root The root element or node to start traversing on.
+   * @return {NodeIterator} The created NodeIterator
    */
-  const _createIterator = function (root) {
+  const _createNodeIterator = function (root) {
```
ssi02014, line 908: https://github.com/microsoft/TypeScript/blob/main/src/lib/dom.generated.d.ts#L7160
I've modified the jsdoc with reference to that documentation.

Also, the naming seems clearer for createNodeIterator than createIterator.

src/purify.js
```diff
@@ -1183,13 +1182,14 @@ function createDOMPurify(window = getGlobal()) {
...
    * checks if at least one dash is included in tagName, and it's not the first char
    * for more sophisticated checking see https://github.com/sindresorhus/validate-element-name
    *
    * @param {string} tagName name of the tag of the node to sanitize
+   * @returns {boolean} Returns true if the tag name meets the basic criteria for a custom element, otherwise false.
    */
-  const _basicCustomElementTest = function (tagName) {
+  const _isBasicCustomElementCheck = function (tagName) {
     return tagName.indexOf('-') > 0;
   };
```
ssi02014, line 1194: I added the prefix `is` because the function differs from the function name in the `jsdoc`, and because it returns a boolean.

jsdoc: _basicCustomElementCheck -> `This seems to fit the intent better.`
actual:  _basicCustomElementTest

src/purify.js
```diff
@@ -900,19 +900,18 @@ function createDOMPurify(window = getGlobal()) {
...
    */
-  const _createIterator = function (root) {
+  const _createNodeIterator = function (root) {
     return createNodeIterator.call(
       root.ownerDocument || root,
       root,
       // eslint-disable-next-line no-bitwise
       NodeFilter.SHOW_ELEMENT | NodeFilter.SHOW_COMMENT | NodeFilter.SHOW_TEXT,
-      null,
-      false
```
ssi02014, line 915: I removed that argument.

src/purify.js
```diff
@@ -900,19 +900,18 @@ function createDOMPurify(window = getGlobal()) {
...
+      null
```
ssi02014, line 914: I removed that argument.

src/purify.js
```diff
@@ -1183,13 +1182,14 @@ function createDOMPurify(window = getGlobal()) {
...
-   * _basicCustomElementCheck
+   * _isBasicCustomElement
    * checks if at least one dash is included in tagName, and it's not the first char
    * for more sophisticated checking see https://github.com/sindresorhus/validate-element-name
    *
    * @param {string} tagName name of the tag of the node to sanitize
+   * @returns {boolean} Returns true if the tag name meets the basic criteria for a custom element, otherwise false.
    */
-  const _basicCustomElementTest = function (tagName) {
+  const _isBasicCustomElement = function (tagName) {
```
ssi02014, line 1192: The function name in the jsdoc was different from the actual function name, so I tried to modify it, and finally changed it to _isBasicCustomElement with the is prefix because it returns a boolean.

jsdoc: _basicCustomElementCheck
actual: _basicCustomElementTest

cure53:
This looks good me thinks, but let me do some manual testing - high risk change :sweat_smile:

ssi02014:
@cure53 
Based on the specification, I don't think the last argument, "false", will make any difference in behavior, but it's worth checking.

Thanks!!!

PR #859, opened 2023-09-15
Update attrs.js for compare

Wind-info:
Fix stuck with other user can make

## Tasks

- xxxx
- xxxx
- xxxx

## Dependencies

- [x] Resolved dependency
- [ ] Open dependency


cure53:
This looks broken, cannot merge @Wind-info

PR #874, opened 2023-11-09
chore: packagejson conditional exports

ssi02014:
## Summary

@cure53 👋
Even if you currently import DOMPurify with import, it will import files in `umd` format.

```tsx
// react env
import DOMPurify from "dompurify";
```
스크린샷 2023-11-09 오후 11 35 33

We can apply `Conditional Exports` so that we get files in `esm` format when importing with import, and files in `cjs` format when importing with require. 

https://nodejs.org/api/packages.html#conditional-exports

Also, why not change it to a `.mjs` file extension to make it clear that it is an `esm` format?

<hr />

It seems like a lot of file changes, but I reinstalled the dependencies and built, and there are differences like line breaks. 😂

스크린샷 2023-11-09 오후 11 43 45


cure53:
Oh, nice, thanks! Just wondering for version purposes, will this break things for folks after a release?

package.json
```diff
@@ -10,17 +10,24 @@
...
-  "main": "dist/purify.cjs.js",
-  "module": "dist/purify.es.js",
-  "browser": "dist/purify.js",
-  "production": "dist/purify.min.js",
+  "exports": {
+    ".": {
+      "require": "./dist/purify.cjs.js",
+      "import": "./dist/purify.es.mjs"
+    },
+    "./package.json": "./package.json"
```
ssi02014, line 25: Many libraries provide package meta information via the following code 👍
```tsx
import packageInfo from "@suspensive/react/package.json";
```

ssi02014:
@cure53 
I don't think this changes much.
If you use `import`, you'll get `"./node_modules/dompurify/dist/purify.es.mjs"`, and if you use `require`, you'll get `"./node_modules/dompurify/dist/purify.cjs.js"`.

If the build files we provide are good, then everything should be fine.

Several other libraries provide this ! 

Here's an example of another library (html-react-parser)
https://github.com/remarkablemark/html-react-parser/blob/master/package.json
```json
  "exports": {
    ".": {
      "types": "./lib/index.d.ts",
      "import": "./esm/index.mjs",
      "require": "./lib/index.js"
    },
    "./lib/*": {
      "types": "./lib/*.d.ts",
      "import": "./esm/*.mjs",
      "default": "./lib/*.js"
    }
  },
```
스크린샷 2023-11-09 오후 11 59 15

or
[Tanstack Query](https://github.com/TanStack/query/blob/main/packages/react-query/package.json)

cure53:
Awesome, thank you very much :bow: :smile:

Issue #895, opened 2024-01-04
[bug] - Build error in angular app


mathiasdeoliveira:
**[bug] - Build error in angular app since your last version 3.0.7**

Error: export 'sanitize' (imported as 'DOMPurify') was not found in 'dompurify' (possible exports: default)

### Background & Context

We use your library with Angular version 15 and dompurify import in non-fixed mode. 

_package.json_
`"dompurify: "^3.0.6",`

and import like this into the typescript file

`import * as DOMPurify from 'dompurify';`

**Temporary solution :** 

use fixed version "dompurify: "3.0.6"

cure53:
Oha, we didnÄt change much that might impact how to import the library, do you have more info per chance?

cure53:
Maybe the changes done in early November are the culprit? cc @ssi02014 

https://github.com/cure53/DOMPurify/compare/3.0.6...3.0.7

Justxs:
I have same issue using React

pauly92:
I have the same using LIT.

This is the error message i get:

Import "sanitize" will always be undefined because there is no matching export in "../../node_modules/dompurify/dist/purify.es.mjs" [import-is-undefined]

cure53:
@ssi02014 If possible, please have a look. I am pretty sure this is because of the changes in November:

https://github.com/cure53/DOMPurify/pull/874#issuecomment-1803975568

CaptainGlac1er:
Also getting this error once my app updated from 3.0.6 -> 3.0.7

ssi02014:
@cure53 
Until v3.0.6, we loaded "dist/purify.js" in `umd` format, but since v3.0.7, we load "dist/purify.es.mjs" in `esm` format, which seems to cause related issues.

Currently, purfiy is exporting via `export default createDOMPurify();` so importing with `import DOMPurify from 'dompurify'"`doesn't seem to cause the issue.

CaptainGlac1er:
@ssi02014 I have the issue importing with `import * as DOMPurify from 'dompurify';`

ssi02014:
```tsx
import DOMPurify from 'dompurify'"
```
Can you tell if you're getting an error if you proceed as above?

cure53:
> ```tsx
> import DOMPurify from 'dompurify'
> ```
> 
> Do you get an error if you do the above?

[...]

PR #896, opened 2024-01-04
fix: revert conditional exports

ssi02014:
## Summary

@cure53 
https://github.com/cure53/DOMPurify/issues/895
I reverted `conditional exports`, which I suspect is the cause of the above issue. 

### Related Pull Request
https://github.com/cure53/DOMPurify/pull/874/files


cure53:
Thanks :smile: Will check an release tomorrow.

ssi02014:
@cure53 
Thank you. 🙏

Personally, I think it is appropriate to import the `esm` format when "import", and it is appropriate to import it like below.
```tsx
import DOMPurify from "dompurify"
```
However, some people are importing `* as DOMPurify`, so I hope that fixes the issue.

Issue #897, opened 2024-01-04
TypeError thrown by clone when configuration contains null values


bob-difronzo:
### Background & Context

When upgrading from `3.0.6` to `3.0.7`, a `TypeError` is getting thrown when calling `sanitize` if the configuration object contains any `null` values. The issue appears to be [this line](https://github.com/cure53/DOMPurify/blob/main/src/utils.js#L139) in the `clone` utility since `typeof null === 'object'`. For now we have replaced `null` with `undefined`.

#### Input

```js
DOMPurify.sanitize(anyString, {
  CUSTOM_ELEMENT_HANDLING: {
    tagNameCheck: null
  }
});
```

#### Given output

```
Uncaught TypeError: Cannot read properties of null (reading 'constructor')
```

#### Expected output

Expected no error to be thrown.

cure53:
Thanks for spotting this and thanks for sending over a PR. The PR looks great and got merged :smile:

PR #898, opened 2024-01-05
fix: Avoid TypeError for null values on CUSTOM_ELEMENT_HANDLING config.

cpmotion:
ISSUE-897

## Summary

[See issue](https://github.com/cure53/DOMPurify/issues/897). Change made in utils and test added.

## Tasks

Not sure if there's a better way to inverse assert in the test. Please advise if so. Tested locally to pass and fail. It "works" but kind of a workaround for QUnit it seems...


cpmotion:
Just realized saving in test-suite applied prettier fixes. I guess prettier is not running in pipeline to validate against config. Hmm, I'm going to undo those lines for this specific commit and update pr..

cure53:
This is great, thanks you!

cpmotion:
@cure53 any chance we could push a patch(3.0.8) out asap? We have a specific scenario where we're picking up latest v3 and that would really help a currently deployed product version. Thank you!

cure53:
This release should fix it: https://github.com/cure53/DOMPurify/releases/tag/3.0.8

cpmotion:
Awesome thanks!

PR #903, opened 2024-01-17
console warn fallback for <b> when run with Happydom

HugoPoi:
## Summary

console.warn(element) is displaying a lot of stuff, in some context like HappyDom this displaying a lot of stuff because a recursive console output of the entire Dom.

## Background & Context

I was thinking of the other solutions
* Change the way of displaying a element in Happydom
* Remove the display of the element in the warn

## Tasks

- Could someone give me some context when this fallback function is called
  `fallback value for <b>by SMS</b>` when calling `DOMPurify.sanitize('The patient will receive a link <b>by SMS</b> to join the video consultation')` so why `<b>by SMS</b>` will be an element considered as a fallback ?

https://github.com/cure53/DOMPurify/blob/756cec35240cefe261aab06be1e0a7a1e66a6b19/src/purify.js#L114-L120

Could be Happydom not implementing the same way the API for `<b>` ?

As now it seems Dompurify doesn't support Happydom so might be an upstream issue in Happydom itself.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -175,8 +175,7 @@ function lookupGetter(object, prop) {
     object = getPrototypeOf(object);
   }
 
-  function fallbackValue(element) {
-    console.warn('fallback value for', element);
+  function fallbackValue() {
     return null;
   }
 


cure53:
Hey there, thanks for sending over the PR. Indeed, we believe, this is an issue in Happydom itself, here is more context:

https://github.com/cure53/DOMPurify/pull/878
https://github.com/capricorn86/happy-dom/issues/1165

I am hesitant to accept the PR since working with outerHTML doesn't really fix this and also adds additional complications, such as for elements that don't have an outerHTML etc.

I am not sure how active Happydom is, but I guess the best fix would be to address the root cause.

cure53:
I am wondering, do we even need the `console.warn` here? @vlad-borodaev 

https://github.com/cure53/DOMPurify/blame/756cec35240cefe261aab06be1e0a7a1e66a6b19/src/utils.js#L179

HugoPoi:
Ok thanks a lot to respond so quickly ! After re-thinking this, we're facing 3 issues here that's are not directly related to Dompurify

1. As you mention the fallback occur because Happydom doesn't implement some part of the API https://github.com/capricorn86/happy-dom/issues/1165
2. Second I need to open issue in Happydom because `console.log(element)` lead to recursively printing earth, not sure we want that
3. Using `console` is a side effect functionally speaking, we probably want to either throw or remove the warn or inject a logger ?

cure53:
I hear you :smile: I want to get rid of the `console.warn` call entirely as well, hence my ping to the one who added it.

Would you mind reshaping the PR so we nuke the whole `console.warn` call? Then we can take it from there.

cure53:
Thanks :bow:

cure53:
In case that works and keeps Happydom happy, we could consider a release soon.

vlad-borodaev:
> I am wondering, do we even need the `console.warn` here? @vlad-borodaev
> 
> https://github.com/cure53/DOMPurify/blame/756cec35240cefe261aab06be1e0a7a1e66a6b19/src/utils.js#L179

Hi everyone, sorry for missing the message. Thanks for the fix, I suppose as it was a fallback to fix the issue with nullable element for Angular Universal on the server side (for Angular 11-12 as I remember), we could definitely remove the console there.

Probably it would be better then to leave a TODO/FIXME-comment to later review this fallback, probably we need to process more cases there.

PR #908, opened 2024-01-29
fix(utils): Fixed getOwnPropertyDescriptor to hasOwnProperty function

ssi02014:
## Summary

@cure53 👋

The [getOwnPropertyDescriptor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptor) static method returns an object describing the configuration of a specific property on a given object (that is, one directly present on an object and not in the object's prototype chain).

My understanding is that `getOwnPropertyDescriptor` within the `cleanArray` and `clone` functions is simply used to validate that the property exists.  (Instead, the `lookupGetter` function actually uses the descriptor)

Therefore, I thought it would be more appropriate to utilize the [hasOwnProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty) function for the purpose of determining if a property exists. (Returns a `boolean` that suits our intent.)

```js
const objectHasOwnProperty = unapply(Object.prototype.hasOwnProperty);

const arr = [1, 2];

console.log(objectHasOwnProperty(arr, 0)); // true
console.log(objectHasOwnProperty(arr, 1)); // true
console.log(objectHasOwnProperty(arr, 2)); // false

const obj = {
  foo: 1,
};

console.log(objectHasOwnProperty(obj, "foo")); // true
console.log(objectHasOwnProperty(obj, "bar")); // false
```

<br />


src/utils.js
```diff
@@ -115,7 +115,9 @@ function addToSet(set, array, transformCaseFunc = stringToLowerCase) {
  */
 function cleanArray(array) {
   for (let index = 0; index < array.length; index++) {
-    if (getOwnPropertyDescriptor(array, index) === undefined) {
+    const isHasOwnProperty = Object.prototype.hasOwnProperty.call(array, index);
+
```
ssi02014, line 119: JavaScript does not protect the property name hasOwnProperty, so objects with properties with this name may return incorrect results.

[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty#using_hasownproperty_as_a_property_name)
[eslint no-prototype-builtins](https://eslint.org/docs/latest/rules/no-prototype-builtins)

src/utils.js
```diff
@@ -115,7 +115,12 @@ function addToSet(set, array, transformCaseFunc = stringToLowerCase) {
  */
 function cleanArray(array) {
   for (let index = 0; index < array.length; index++) {
-    if (getOwnPropertyDescriptor(array, index) === undefined) {
+    const hasOwnPropertyCheck = Object.prototype.hasOwnProperty.call(
+      array,
+      index
+    );
```
ssi02014, line 121: JavaScript does not protect the property name hasOwnProperty, so objects with properties with this name may return incorrect results.

I think this is a stable way to use it.

Related
[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty#using_hasownproperty_as_a_property_name)
[eslint no-prototype-builtins](https://eslint.org/docs/latest/rules/no-prototype-builtins)

src/utils.js
```diff
@@ -115,7 +115,9 @@ function addToSet(set, array, transformCaseFunc = stringToLowerCase) {
  */
 function cleanArray(array) {
   for (let index = 0; index < array.length; index++) {
-    if (getOwnPropertyDescriptor(array, index) === undefined) {
+    const isPropertyExist = Object.prototype.hasOwnProperty.call(array, index);
```
ssi02014, line 118: JavaScript does not protect the property name hasOwnProperty, so objects with properties with this name may return incorrect results.

I think this is a stable way to use it.

#### Related
[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty#using_hasownproperty_as_a_property_name)
[eslint no-prototype-builtins](https://eslint.org/docs/latest/rules/no-prototype-builtins)

src/utils.js
```diff
@@ -133,7 +135,12 @@ function clone(object) {
   const newObject = create(null);
 
   for (const [property, value] of entries(object)) {
-    if (getOwnPropertyDescriptor(object, property) !== undefined) {
+    const isPropertyExist = Object.prototype.hasOwnProperty.call(
+      object,
+      property
+    );
```
ssi02014, line 141: JavaScript does not protect the property name `hasOwnProperty`, so objects with properties with this name may return incorrect results.

This is a more stable way to use it.

#### Related
[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty#using_hasownproperty_as_a_property_name)
[eslint no-prototype-builtins](https://eslint.org/docs/latest/rules/no-prototype-builtins)

src/utils.js
```diff
@@ -46,6 +46,8 @@ const stringReplace = unapply(String.prototype.replace);
 const stringIndexOf = unapply(String.prototype.indexOf);
 const stringTrim = unapply(String.prototype.trim);
 
+const objectHasOwnProperty = unapply(Object.prototype.hasOwnProperty);
```
ssi02014, line 49: JavaScript does not protect the property name `hasOwnProperty`, so objects with properties with this name may return incorrect results.

This is a more stable way to use it.

### Related
[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty#using_hasownproperty_as_a_property_name)
[eslint no-prototype-builtins](https://eslint.org/docs/latest/rules/no-prototype-builtins)

PR #910, opened 2024-02-15
fix: Change the "in" operator to hasOwnProperty

ssi02014:
## Summary
@cure53 👋
Related: https://github.com/cure53/DOMPurify/pull/908

Why not use the `hasOwnProperty` function to overcome the problems with the `in` operator?
In practice, this won't change much of the behavior. However, I expect it would be more stable..!! 🙏

## Reference
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in#inherited_properties
https://stackoverflow.com/questions/13632999/if-key-in-object-or-ifobject-hasownpropertykey

## Dependencies

- [x] Resolved dependency
- [x] Open dependency


cure53:
Looks good to me, thank you again :slightly_smiling_face:

PR #942, opened 2024-04-25
docs(README.md): correct hook name in example and remove misleading comment

kyselberg:
## Summary

this pull request changes hook usage example 

## Background & Context

now hookEvent is passed **only** for `uponSanitizeAttribute` hook.
currently there is misleading example of hook usage, where the `beforeSanitizeElements` hook is used.

in order to avoid misunderstandings and simplify life for next users of lib, let's correct this mistake.


cure53:
Oh, well spotted, thank you!

Issue #526, opened 2021-03-22
DOMPurify.sanitize is not a function


akashgp09:
### Background & Context

I am using **DOMPurify** in my react App

### Bug
**DOMPurify.sanitize is not a function**

#### Input
```
import DOMPurify from "dompurify";

const dirty = `I love to do evil <img src="http://unsplash.it/100/100?random" onload="alert('you got hacked');" />`;

const App = () => (
  <div
    dangerouslySetInnerHTML={{
      __html: DOMPurify.sanitize(dirty)
    }}
  />
);
```
#### Given output

**DOMpurify.sanitize is not a function** 

#### Expected output

Should execute well

cure53:
Not our bug, me thinks.

binyamin:
@cure53 same problem here. Do you have any idea who's bug it _is_?

binyamin:
Ah, maybe it needs jsdom? The readme doesn't seem to say that in node.js, you need jsdom to work

Fensterbank:
I had this issue with NextJS on server-side rendering.
To make this work server-side and client-side the same way, you can use the wrapper https://github.com/kkomelin/isomorphic-dompurify

duoluodexiaoxiaoyuan:
> ### Background & Context
> I am using **DOMPurify** in my react App
> 
> ### Bug
> **DOMPurify.sanitize is not a function**
> 
> #### Input
> ```
> import DOMPurify from "dompurify";
> 
> const dirty = `I love to do evil <img src="http://unsplash.it/100/100?random" onload="alert('you got hacked');" />`;
> 
> const App = () => (
>   <div
>     dangerouslySetInnerHTML={{
>       __html: DOMPurify.sanitize(dirty)
>     }}
>   />
> );
> ```
> 
> #### Given output
> **DOMpurify.sanitize is not a function**
> 
> #### Expected output
> Should execute well

i  find a question , this is a demo https://codesandbox.io/s/ei98d?file=/src/App.tsx

s1nistr4:
Eh, if the dev is gonna not fix the bug and also be rude when people bring it up, I recommend people check out sanitize-html instead. Easy fix, works great.

bonface221:
🤔

mtoha2013:
I have similar issue, how to achieve this at pure javascript no framework like react?
I run it on .NET MVC app

Bunpasi:
[...]

PR #956, opened 2024-05-15
Documentation

MortenHofft:
https://github.com/cure53/DOMPurify/issues/526#issuecomment-2112543369

## Summary

The example doesn't work, it is described in https://github.com/cure53/DOMPurify/issues/526#issuecomment-1435711907 what it ought to be. This is simply fixing the typo


cure53:
Oh, nice, thanks a lot!

PR #957, opened 2024-05-17
Allow Popover API attributes

Gigabyte5671:
## Summary

This PR adds the new attributes from the [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API) to the list of allowed HTML attributes. This includes:
- `popover`
- `popovertarget`
- `popovertargetaction`

These attributes are used to identify popover elements and control their behaviour. I don't believe they can be used to trigger scripts in any way, but please let me know if I'm wrong.

Thanks!

## Background & Context

This API is quite new, but support for it is already estimated at ~84% ([caniuse](https://caniuse.com/?search=popover)).


cure53:
Thanks, we should be fine with those as far as I can see, time will tell.

PR #972, opened 2024-06-17
feat(docs): add removed options

Rotzbua:
## Summary

Document removed options.

## Background & Context

Make it easier to migrate from older versions.

## Tasks

- Are there other removed options?

## Dependencies

- [x] Resolved dependency
- [ ] Open dependency


cure53:
This is a great idea, thanks!

PR #990, opened 2024-09-03
Added support for the Angular compiler.

### LINKED ISSUE #869: Trying to use in Angular 16, getting "CommonJS or AMD dependencies can cause optimization bailouts."
Is it possible to have an ES6 module export version of this?

jeroen1602:
## Summary

The Angular compiler looks for the main field using a few fields in the package.json. The order it uses is:
- es2020
- es2015
- browser
- module
- main

The first entry the Angular compiler would find before this change was the 'browser' entry point. This is however compiled as an `umd` and Angular is looking for an `esm` so it would complain (it would still work though). Adding `es2020` as an entry point fixes this problem.

Fixes #869

## Tasks

1. Create an Angular application 
2. Add this package as a dependency `npm install --save ./path/to/project`
3. Import `DOMPurify` in the `AppComponent` and just sanitize some string in the `ngOnInit` function
4. Run `npm run build` to build this application and see that the warning about common js modules is resolved.

## Dependencies

- [x] Resolved dependency
- [ ] Open dependency


jeroen1602:
I don't known how I did it but somehow I changed more files than I wanted, will try to amend this commit.

cure53:
Most excellent, thank you :)

PR #1008, opened 2024-10-22
fix(types): export `Config` type from package

aloisklink:
## Summary

This PR exports the `Config` type from DOMPurify, so that people can do:

```ts
import type {Config} from 'dompurify';

// or
type Config = import('dompurify').Config;
```

## Background & Context

This is to more closely match [@types/dompurify][1], which allows it, to make migration easier for any TypeScript users that currently use `@types/dompurify`.

[1]: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a0058c8132907c88b6e4d3dd0f4103f799deb732/types/dompurify/index.d.ts#L59

## Tasks

N/A

## Dependencies

N/A


cure53:
This looks great, thank you :)

PR #1011, opened 2024-10-27
Maintaining the order of the attributes in sanitized html.

parveshgoyal01:
## Summary

Maintaining the order of the attributes in sanitized html.

## Background & Context

After every sanitize, the output keeps on changing.

For instance,
when input given is `<div class="column-2" style="padding-top: 10px;"></div>`

sanitized output comes out to be : `<div style="padding-top: 10px;" class="column-2"></div>`

and
when input given is `<div style="padding-top: 10px;" class="column-2"></div>`

sanitized output comes out to be : `<div class="column-2" style="padding-top: 10px;"></div>`

Because of above, output is not consistent across multiple sanitizations.


cure53:
Sadly we cannot allow for this one as it causes security issues and bypasses, sorry.

PR #1019, opened 2024-11-13
Update CUSTOM_ELEMENT regexp to support custom elements with colon into the tag.

ciccio-90:
I need to support custom elements that have colon into the tag definition. Example:
<app:custom-element>...</app:custom-element>

### DIFF
--- a/src/regexp.ts
+++ b/src/regexp.ts
@@ -14,4 +14,4 @@ export const ATTR_WHITESPACE = seal(
   /[\u0000-\u0020\u00A0\u1680\u180E\u2000-\u2029\u205F\u3000]/g // eslint-disable-line no-control-regex
 );
 export const DOCTYPE_NAME = seal(/^html$/i);
-export const CUSTOM_ELEMENT = seal(/^[a-z][.\w]*(-[.\w]+)+$/i);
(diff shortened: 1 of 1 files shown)


ciccio-90:
No, it cannot because the problem is into the method _isBasicCustomElement that checks if the custom element tag matches the regular expression provided by CUSTOM_ELEMENT constant.

cure53:
I see I see, well - then we need to find a different solution :slightly_smiling_face: Changing a core regex with the risk of enabling XSS via namespaces is not an option, sorry.

ciccio-90:
Ok can you provide me a different solution please?

cure53:
No :slightly_smiling_face: not my problem to solve. 

I would check if that can be handled with a hook for example, that might be the way to go for weird tags.

ciccio-90:
What do you mean handled with a hook? An existing one? Or you will provide a new one?

cure53:
Check here: https://github.com/cure53/DOMPurify#hooks

You should be able to intercept the sanitization in case a colon is found in the element name, then react, then continue. This should be the way to go without patching core regex and potentially causing bypasses.

PR #1022, opened 2024-11-14
fix: ignore export assign cannot be used in module TS error, fixes #1018

### LINKED ISSUE #1018: 3.2.0 seems to cause a build error in Angular/React: An export assignment cannot be used in a module with other exported elements.
### Background & Context

I tried to upgrade to latest v3.2.0 that drops the need for the `@types/dompurify` which is awesome, however my Angular build is now failing with this error (you can see the error in my CI workflow [failure](https://github.com/ghiscoding/Angular-Slickgrid/actions/runs/11803253896/job/32880801702))

> Error: node_modules/dompurify/dist/purify.cjs.d.ts:401:1 - error TS2309: An export assignment cannot be used in a module with other exported elements.
> 
> 401 export = _default;

If we browse to the CJS dist bundle from [unpkg](https://unpkg.com/browse/dompurify@3.2.0/dist/purify.cjs.d.ts), we can see that it's the last line of the file export that it doesn't seem to like. I don't fully understand the problem but it might be due to trying to export DOMPurify twice? Though again I'm not entirely sure, see this Stack Overflow [answer](https://stackoverflow.com/a/48708568/1212166) for that error

### Bug

Also another possible issue that I found on top of what is described above, is that I was a bit surprised that it picked up the CJS (CommonJS) build instead of the expect ESM dist. Taking a look at the `package.json`, it is most probably because the CJS `require` is the first export to appear, however it should really be the `import` that comes first and make the `require` a fallback (2nd) which is the inverse of what is currently set. Note that I could provide a PR for the correct order but I'm not really that will fix the 1st error above

https://github.com/cure53/DOMPurify/blob/2cf6e2510699261a0004e3431c4e8a0bbd65b230/package.json#L29-L38

We can see the correct order in NodeJS [conditional exports](https://nodejs.org/api/packages.html#conditional-exports)

> Within the ["exports"](https://nodejs.org/api/packages.html#exports) object, key order is significant. During condition matching, earlier entries have higher priority and take precedence over later entries. The general rule is that conditions should be from most specific to least specific in object order.

#### Input

n/a

#### Given output

n/a

#### Expected output

no build error

### Feature

n/a

ghiscoding:
fixes #1018

## Summary

Fixes issue #1018 by simply ignoring the TypeScript error since the TS error has no negative effect whatsoever on the functionality of these types. This makes TypeScript and "Are the Types Wrong" both happy and everything is back to normal and we can get back to work (finally).

The quick explanation is that sometime even when using ESM imports, CJS types exports get picked up (depending on the users tsconfig module resolutions) and `export = ...` is a TypeScript error, but we can easily patch it by ignoring it. 

image

## Background & Context

I found this article [Default exports in CommonJS libraries](https://blog.andrewbran.ch/default-exports-in-commonjs-libraries/) which explains why `export = _default` was patched in the first place. However this gives a TypeScript error in some cases (depending on your tsconfig module resolutions and other configs), this error can simply be ignored altogether because the types still work correctly, the error when active though will block a TypeScript build and ignoring the error will let us continue with our build. 

image

## Tasks

Simply modified the existing `fix-cjs-types.js` script, run a build.

This also keeps "Are the Types Wrong" happy since nothing really changed for the types themselves, so all is good.

image

## Dependencies

- [x] Resolved dependency
- [ ] Open dependency

### DIFF
--- a/dist/purify.cjs.d.ts
+++ b/dist/purify.cjs.d.ts
@@ -398,4 +398,5 @@ type WindowLike = Pick<typeof globalThis, 'DocumentFragment' | 'HTMLTemplateElem
 
 export { type Config, type Hook, type HookName, type RemovedAttribute, type RemovedElement, type UponSanitizeAttributeHook, type UponSanitizeAttributeHookEvent, type UponSanitizeElementHook, type UponSanitizeElementHookEvent, type WindowLike };
 
+// @ts-ignore
 export = _default;
--- a/scripts/fix-cjs-types.js
+++ b/scripts/fix-cjs-types.js
@@ -22,7 +22,7 @@ const path = require('node:path');
   // Append `export = _default;` to match the
   // `module.exports = DOMPurify` statement
   // that Rollup creates.
-  fixed += '\nexport = _default;\n';
+  fixed += '\n// @ts-ignore\nexport = _default;\n';
 
   await fs.writeFile(fileName, fixed);
 })().catch((ex) => {


ghiscoding:
@reduckted could you review the PR?

reduckted:
Looks good. Thanks for investigating and finding a fix! ❤️

cure53:
Thanks everyone!

PR #1024, opened 2024-11-14
Ensured that purify.min.js is actually minified.

reduckted:
## Summary

This change ensures that the `purify.min.js` file is actually minified.

## Background & Context

The minified UMD output was not being minified because it relied on the npm script being called `build`. The actual npm script that runs rollout is called `build:rollup` (and maybe something different from tests?). In any case, I've removed the conditional so that the "production" output is always minified.

Thank you @bpc-clone for spotting this.

## Tasks

N/A

## Dependencies

N/A


cure53:
Thank you very much again :)

PR #1031, opened 2024-11-20
Changed the type of the "node" in hooks to be as specific as possible

### LINKED ISSUE #1029: Incorrect type for addHook callback currentNode argument
> This issue proposes a [bug, feature] which... Bug

### Background & Context

`addHook` callback now typed with `Node` instead of `Element`

### Bug

In the latest version (3.0.5) of the DefinitelyTyped types for the `addHook` method, the `currentNode` argument to the callback is typed as `Element`.

```ts
// @types/dompurify/index.d.ts

addHook(hook: HookName, cb: (currentNode: Element, data: HookEvent, config: Config) => void): void;
```

In the first-party types bundled with the 3.2.0 release, this changes to `Node`

```ts
// DOMPurify/dist/purify.cjs.d.ts
...
addHook(entryPoint: BasicHookName, hookFunction: Hook): void;
...
type Hook = (this: DOMPurify, currentNode: Node, hookEvent: null, config: Config) => void;
```
 
Its possible this was intentional since `Element` extends `Node` and may not be appropriate for all hooks; however, in doing so, you have broken one of your own examples.  https://github.com/cure53/DOMPurify/blob/f93acffc37667f7e6a008d7396c961beba831c92/demos/hooks-target-blank-demo.html

```ts
// Add a hook to make all links open a new window
DOMPurify.addHook('afterSanitizeAttributes', (node) => {  // node is now typed as Node, not Element
    if ('target' in node) {
        node.setAttribute('target', '_blank');  // property setAttribute does not exist on type Node
        node.setAttribute('rel', 'noopener noreferrer');
    }
    if (!node.hasAttribute('target') && (node.hasAttribute('xlink:href') || node.hasAttribute('href'))) {  // property hasAttribute does not exist on type Node
       node.setAttribute('xlink:show', 'new');
    }
});
```

`setAttribute` and `getAttribute` are implemented on `Element`, not `Node`, so this example no longer satisfies TypeScript.

Would like some clarification on this. Was this change intentional? If so, that means that hooks must be written to support general `Node` objects and not just `Element` objects. Do you have an updated example of setting the target/rel attributes in a link element?

reduckted:
Fixes #1029.

Prior to the TypeScript type declarations being provided by this package, the `@types/dompurify` package defined all hooks as taking an `Element` as the first parameter. That was incorrect because some hooks can be given a `Node` or `DocumentFragment`. When the type declarations were moved into this repository, the hook's first parameter type was changed to `Node`. This change was overzealous because some hooks are guaranteed to only be given an `Element` or `DocumentFragment`.

To account for those hooks, this pull request adds two additional hook types - `ElementHook` and `DocumentFragmentHook` (plus `Hook` has been renamed to `NodeHook` to disambiguate it). 
|Hook| Type of `currentNode`|
|----|----|
|`beforeSanitizeElements`| `Node`|
| `afterSanitizeElements`| `Node`|
| `beforeSanitizeShadowDOM`| `DocumentFragment`|
| `uponSanitizeShadowNode`| `Node`|
| `afterSanitizeShadowDOM`|`DocumentFragment`|
| `beforeSanitizeAttributes`| `Element`|
| `afterSanitizeAttributes`| `Element`|
| `uponSanitizeElement`| `Node`|
| `uponSanitizeAttribute`| `Element`|

To make this as type-safe as possible internally, I've changed the hooks to be stored in an object with properties named after each hook rather than just in an object that keys a string to an array of functions. The `_executeHooks` function now takes in the array of hooks to execute rather than the name of the hook. This allows the "node" and "data" parameters for the hook to have type-checking applied at compile time to ensure that, for example, you don't call `_executeHooks` with a `Node` when the hooks you are executing require an `Element`.


cure53:
Awesome, thank you!

PR #1036, opened 2024-11-26
Add triple-slash types reference to trusted-types.

### LINKED ISSUE #1035: Compile of Angular app not possible since update to v3.2.x
### Background & Context

We tried to update from v3.17 to v3.20 or v3.2.1, but the compile of our Angular app fails. Installing @types/trusted-types 2.0.7 does not change anything.

### Bug

When running the build. the following error message appears:

```
> ng-web@3.5.0 build:app
> ng build in-web-app --configuration production --aot

✔ Browser application bundle generation complete.

Error: node_modules/dompurify/dist/purify.es.d.mts:170:28 - error TS2304: Cannot find name 'TrustedTypePolicy'.

170     TRUSTED_TYPES_POLICY?: TrustedTypePolicy | undefined;
                               ~~~~~~~~~~~~~~~~~


Error: node_modules/dompurify/dist/purify.es.d.mts:245:9 - error TS2304: Cannot find name 'TrustedHTML'.

245     }): TrustedHTML;
            ~~~~~~~~~~~


Error: node_modules/dompurify/dist/purify.es.d.mts:396:34 - error TS2339: Property 'trustedTypes' does not exist on type 'Window & typeof globalThis'.

396     trustedTypes?: typeof window.trustedTypes;
                                     ~~~~~~~~~~~~

```

#### Input

n/a

#### Given output

n/a

#### Expected output

no build error

### Feature

n/a

reduckted:
## Summary

Fixes #1035.

We need to tell TypeScript that we use the type declarations from `trusted-types`. This is done using a [triple-slash directive](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-types-). Without this, the references to `trusted-types` in the type declaration file can cause compilation errors for certain configurations.

I initially added `/// <reference types="trusted-types" preserve="true"/>` to the top of `purify.ts`. Doing that resulted in the triple-slash directive being included in the type declaration files as expected, _but_ it also ended up in the compiled JavaScript, but not at the top of the file (I think due to Rollup combining files together). To keep those files clean I've opted to add this directive via the `fix-types.js` script (previously called `fix-cjs-types.js`).

I've also installed `@types/node` as a dev dependency so that `fix-types.js` can have type checking performed to help avoid any mistakes when editing that file.


cure53:
Thanks a ton, again! :bow:

PR #1037, opened 2024-11-26
Fix #1033

### LINKED ISSUE #1033: Small syntax enhancement to TMPLIT_EXPR
> This issue proposes a small enhancement to the regexp in the `TMPLIT_EXPR` const.

### Background & Context

`const TMPLIT_EXPR = seal(/\${[\w\W]*}/gm);`

Projects that use XSLT can inline 3rd party libs before pushing out the response to the browser. While doing so they can perform code resolution/substitutions on ${} espressions. Unfortunately the `${[\w\W]*}` part can be misinterpreted as an expression even though it's in a regexp, and because it obviously doesn't resolve to anything during the transformation, it gets wiped out as a result and causes a syntax error when the browser gets it.
Result: `const TMPLIT_EXPR = seal(/\/gm);`

### Proposed workaround

The solution is extremely simple and doesn't change anything to how the regexp operates, and would improve compatibility with a wider range of mission critical projects build on "older" tech like XSLT.
All that is needed is to escape the first curly brace like so:
`const TMPLIT_EXPR = seal(/\$\{[\w\W]*}/gm);`

Would you like me to create a PR for this?

svdb99:
## Summary

Small fix for issue #1033 

## Background & Context

`const TMPLIT_EXPR = seal(/\${[\w\W]*}/gm);`

Projects that use XSLT can inline 3rd party libs before pushing out the response to the browser. While doing so they can perform code resolution/substitutions on ${} espressions. Unfortunately the `${[\w\W]*}` part can be misinterpreted as an expression even though it's in a regexp, and because it obviously doesn't resolve to anything during the transformation, it gets wiped out as a result and causes a syntax error when the browser gets it.
Result: `const TMPLIT_EXPR = seal(/\/gm);`

## Proposed solution

The solution is extremely simple and doesn't change anything to how the regexp operates as it merely escapes the first curly brace.


svdb99:
The Lint error wants to optimize the regexp by removing the escape character I added.
Well, no. This "optimization" is precisely what caused the issue in the first place.

cure53:
Feel free to silence it with a eslint-disable :D

cure53:
We are planning a release later today, are you planning to get this change in and amend the PR? :)

svdb99:
Done.

svdb99:
What's your time zone? I'm in CET (UTC+1)

cure53:
Awesome, thanks :slightly_smiling_face:

cure53:
> What's your time zone? I'm in CET (UTC+1)

Same

cure53:
This is again breaking the tests and we have already invested a lot off useless effort into adding one character, I would really recommend that you don't blindly commit stuff but actually run the linter before.

The comment you want to add is this: `// eslint-disable-line unicorn/better-regex`

cure53:
Thanks, this looks better and tests don't die on init this time. In case all good, we will merge and then  release 3.2.2

svdb99:
Sorry, I'm a security analyst not a software developer nor lint expert, and I'm not allowed to install any of your tools on my laptop, so I did indeed copy the same comment from the next line assuming it would be the same. I usually only test new library versions for security breaches and upgrade them in our products when needed (we use over 1800 different libraries). But for DomPurify it was impossible due to this single character.
Thank you for your understanding.

cure53:
No worries, we managed after all :slightly_smiling_face:

PR #1041, opened 2024-12-04
Removed redundant comment

CoryHrycko:
Removed redundant comment

## Summary

Deleted Redundent comment

## Background & Context

Reading through the code to determine if we should use this library, I found a redundant comment. Removing this for readability. 

## Tasks
Deleted a comment


cure53:
LGTM, thanks :slightly_smiling_face:

PR #1054, opened 2024-12-27
feat: allow specifying a hook with `removeHook()`

davecardwell:
## Summary

This pull request implements an optional second argument to `removeHook()` to allow specifying the hook function to remove.

## Background & Context

Currently you can only pop hooks off the stack.

It’s a pretty straight-forward implementation and non-breaking change to allow for specifying the hook to remove.

Because `DOMPurify`’s hooks are global it is otherwise difficult to write code with multiple paths that may be adding hooks and tidying up after themselves. Now it can be trivially written like:

```js
try {
  DOMPurify.addHook('afterSanitizeElements', myHook);

  // …
} finally {
  DOMPurify.removeHook('afterSanitizeElements', myHook);
}
```

## Tasks

- Review implementation changes
- Confirm new test is sufficient


cure53:
This looks great, thank you very much :)

PR #1059, opened 2025-01-21
fix: handle undefined Element in DOMPurify initialization

yehuya:
## Summary

In some cases, `window.Element` is undefined. As a result, importing the DOMPurify package causes it to crash immediately due to the `const ElementPrototype = Element.prototype;` expression in the DOMPurify initialization.

## Background & Context

This fix addresses the error that occurs when importing DOMPurify in environments where `window.Element` is undefined.

## Tasks

- Validate that I didn't break anything.

## Dependencies

- [x] Resolved dependency
- [ ] Open dependency

### DIFF
--- a/test/test-suite.js
+++ b/test/test-suite.js
@@ -871,15 +871,15 @@
         undefined
       );
       assert.strictEqual(
-        typeof DOMPurify({ ...window, Element: undefined }).version,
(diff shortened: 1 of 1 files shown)


cure53:
Oh, this looks great, thank you. This is ready to merge, no?

PR #1060, opened 2025-01-22
tests: DOMPurify custom window test improvement

yehuya:
## Summary

1. Improved the "DOMPurify custom window" test to ensure that the `document` and `window.Element` are tested separately.
2. Increased the likelihood of the test failing if a new `window` property is required in the future, thereby reducing the risk of future `TypeError: Cannot read properties of undefined` errors.


cure53:
Thank you, this LGTM

PR #1065, opened 2025-02-19
chore(package.json): add `dist/purify.js` and `dist/purify.min.js` to exports as targetable modules

reey:
## Summary

This pull request tries to solve a similar issue as described in: https://github.com/cure53/DOMPurify/issues/1050 and fixed in https://github.com/cure53/DOMPurify/pull/1053

In our case we are importing this lib via `dompurify/dist/purify.js`. Which fails with the recent versions of this lib since the `exports` field was introduced with this error message:
```
Error: Module not found: Error: Package path ./dist/purify.js is not exported from package node_modules/dompurify (see exports field in node_modules/dompurify/package.json)
```

## Background & Context

I won't go to much into details here as it is basically the same story as in https://github.com/cure53/DOMPurify/pull/1053.
This could probably also be solved by anyone using this lib by adapting the import of this lib from `dompurify/dist/purify.js` to `dompurify/purify.js`, but this way it would not require a change for anyone still using `dompurify/dist/purify.js`.

## Tasks

- Install PR
- Import DOMPurify
- Using any of the existing methods should continue to work (ie: DOMPurify.sanitize)

## Dependencies

- [x] Resolved dependency
- [ ] Open dependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,9 @@
       }
     },
     "./purify.min.js": "./dist/purify.min.js",
-    "./purify.js": "./dist/purify.js"
+    "./purify.js": "./dist/purify.js",
+    "./dist/purify.min.js": "./dist/purify.min.js",
+    "./dist/purify.js": "./dist/purify.js"
   },
   "files": [
     "dist"


cure53:
Hey there, thanks a lot, this looks great :slightly_smiling_face:

PR #1075, opened 2025-03-15
test: verify various typescript configurations

reduckted:
## Summary

This pull request adds an npm script called `verify-typescript` that will attempt to compile some TypeScript projects with different configurations that use DOMPurify.

## Background & Context

See #1035 and #1073. It's easy to make a change to the TypeScript declarations and verify that it works for one or more configurations, but end up breaking another configuration. The goal of this change is to have a set of known TypeScript configurations that work. If we discover that another configuration doesn't work, then we can add the configuration to these tests and then go about fixing the TypeScript declarations to support that configuration, while also verifying that whatever change is made doesn't break any other configurations that we know about.

The TypeScript code in each project is not meant to be comprehensive. It is just enough to confirm that you can import from `dompurify` and call the functions. 

- Importing `dompurify` verifies that the export is correct, and that the type declarations for `trusted-types` works alongside `dompurify`.
- Calling the functions verifies that the value imported from `dompurify` is a function that creates a new `DOMPurify` instance, and that it also defines the `sanitize` function.

There are currently seven scenarios:
1. A standard CommonJS build.
2. A CommonJS build that specifies specific `types` to include. Notably, `trusted-types` is not one of them (this was actually what caused https://github.com/cure53/DOMPurify/issues/1035#issuecomment-2497890304
3. A CommonJS build that specifies a `typeRoots` to simulate not having `@types/trusted-types` installed.
4. A standard ESM build.
5. An ESM build that specifies specific `types` to include (and, like the CommonJS one does not include `trusted-types`)
6. An ESM build that specifies a `typeRoots` to simulate not having `@types/trusted-types` installed.
7. A `moduleResolution` of `nodenext`.

All of these projects compile successfully, and I have tried applying the change from #1073 and confirmed that it does result in compilation failures. 🎉 

Example output when all succeed:
```
> npm run verify-typescript

> dompurify@3.2.4 verify-typescript
> node ./typescript/verify.js

Installing node modules...

✔ commonjs   
✔ commonjs-with-no-types   
✔ commonjs-with-specific-types   
✔ esm   
✔ esm-with-no-types   
✔ esm-with-specific-types   
✔ nodenext   
```

Example output when there are failures (in this case I applied the changes from #1073):
```
> npm run verify-typescript

> dompurify@3.2.4 verify-typescript
> node ./typescript/verify.js

Installing node modules...

✔ commonjs   
X commonjs-with-no-types   
     C:/Code/GitHub/cure53/DOMPurify/dist/purify.cjs.d.ts (171,28): Cannot find name 'TrustedTypePolicy'.
     C:/Code/GitHub/cure53/DOMPurify/dist/purify.cjs.d.ts (246,9): Cannot find name 'TrustedHTML'.
     C:/Code/GitHub/cure53/DOMPurify/dist/purify.cjs.d.ts (436,34): Property 'trustedTypes' does not exist on type 'Window & typeof globalThis'.
X commonjs-with-specific-types   
     C:/Code/GitHub/cure53/DOMPurify/dist/purify.cjs.d.ts (171,28): Cannot find name 'TrustedTypePolicy'.
     C:/Code/GitHub/cure53/DOMPurify/dist/purify.cjs.d.ts (246,9): Cannot find name 'TrustedHTML'.
     C:/Code/GitHub/cure53/DOMPurify/dist/purify.cjs.d.ts (436,34): Property 'trustedTypes' does not exist on type 'Window & typeof globalThis'.
✔ esm   
X esm-with-no-types   
     C:/Code/GitHub/cure53/DOMPurify/dist/purify.es.d.mts (171,28): Cannot find name 'TrustedTypePolicy'.
     C:/Code/GitHub/cure53/DOMPurify/dist/purify.es.d.mts (246,9): Cannot find name 'TrustedHTML'.
     C:/Code/GitHub/cure53/DOMPurify/dist/purify.es.d.mts (436,34): Property 'trustedTypes' does not exist on type 'Window & typeof globalThis'.
X esm-with-specific-types   
     C:/Code/GitHub/cure53/DOMPurify/dist/purify.es.d.mts (171,28): Cannot find name 'TrustedTypePolicy'.
     C:/Code/GitHub/cure53/DOMPurify/dist/purify.es.d.mts (246,9): Cannot find name 'TrustedHTML'.
     C:/Code/GitHub/cure53/DOMPurify/dist/purify.es.d.mts (436,34): Property 'trustedTypes' does not exist on type 'Window & typeof globalThis'.
✔ nodenext   
```

## Tasks

N/A

## Dependencies

N/A

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -43,3 +43,5 @@ jobs:
           TEST_PROBE_ONLY: ${{ github.ref != 'refs/heads/main' && github.ref != 'refs/heads/2.x' }}
           BS_USERNAME: ${{ secrets.BS_USERNAME }}
           BS_ACCESSKEY: ${{ secrets.BS_ACCESSKEY }}
+      - name: Verify TypeScript
+        run: npm run verify-typescript
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
     "test:jsdom": "cross-env NODE_ENV=test BABEL_ENV=rollup node test/jsdom-node-runner --dot",
     "test:karma": "cross-env NODE_ENV=test BABEL_ENV=rollup karma start test/karma.conf.js --log-level warn ",
     "test:ci": "cross-env NODE_ENV=test BABEL_ENV=rollup npm run test:jsdom && npm run test:karma -- --log-level error --reporters dots --single-run --shouldTestOnBrowserStack=\"${TEST_BROWSERSTACK}\" --shouldProbeOnly=\"${TEST_PROBE_ONLY}\"",
-    "test": "cross-env NODE_ENV=test BABEL_ENV=rollup npm run lint &&  npm run test:jsdom && npm run test:karma -- --browsers Chrome"
+    "test": "cross-env NODE_ENV=test BABEL_ENV=rollup npm run lint &&  npm run test:jsdom && npm run test:karma -- --browsers Chrome",
+    "verify-typescript": "node ./typescript/verify.js"
   },
   "main": "./dist/purify.cjs.js",
   "module": "./dist/purify.es.mjs",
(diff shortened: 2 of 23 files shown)


cure53:
This is amazing, thank you very much again :smile:

PR #1085, opened 2025-04-11
Remove attribute only when necessary

michalnieruchalski-tiugo:
## Summary
The change is to make `_sanitizeAttributes()` only remove an attribute if needed. The way that attributes are sanitized is not changed.

## Background & Context
This provides a performance benefit as attributes are not constantly removed and added back. It also prevents attribute re-ordering, making comparing the pre-sanitized and post-sanitzed content easier and simplifying writing tests. The engineers at Tiny have had a patch applied to dompurify with these changes in production for many years and would like to contribute the changes upstream.

## Tests
As attributes are not removed and re-added, the original attribute order should be maintained after sanitization. As a result, the order of attributes in several tests has been updated. Nothing else about the test cases has been altered.

## Tasks

* Review my changes :)

## Dependencies

N/A


cure53:
Oha, that looks very interesting, thanks a lot! We'll give it a spin.

If all looks fine, we would release this as 3.3.0 as this might be a breaking change for some users.

Issue #1093, opened 2025-04-25
Update attributeNameCheck to include tagName


nelstrom:
Suppose that I want to support the following custom elements:

1. Valid: `<element-one attribute-one="1"></element-one>`
2. Valid: `<element-two attribute-two="2"></element-two>`
3. Invalid: `<element-one attribute-two="2"></element-one>`
4. Invalid: `<element-two attribute-one="1"></element-two>`

I can get close with this:

```js
{
  CUSTOM_ELEMENT_HANDLING: {
    tagNameCheck: (tagName) => tagName.match(/^element-(one|two)$/),
    attributeNameCheck: (attr) => attr.match(/^attribute-(one|two)$/),
    allowCustomizedBuiltInElements: false,
  },
}
```

This would allow my valid examples to pass, but it would also allow my invalid examples to pass.

If the function signature of the `attributeNameCheck` could receive a second argument, representing the `tagName`, then the configuration could be defined as follows:

```js
{
  CUSTOM_ELEMENT_HANDLING: {
    tagNameCheck: (tagName) => tagName.match(/^element-(one|two)$/),
    attributeNameCheck: (attr, tagName) => {
      if (tagName === 'element-one') {
        return ['attribute-one'].includes(attr);
      } else if (tagName === 'element-two') {
        return ['attribute-two'].includes(attr);
      } else {
        return false;
      }
    },
    allowCustomizedBuiltInElements: false,
  },
}
```

Is this potentially a feature that you'd be willing to support? Are there any security concerns or performance concerns that would make this unacceptable?

nelstrom:
That's good to know @cure53, thanks for your response. I might consider opening a PR for this.

cure53:
Awesome, thanks - let me know please if any help is needed.

PR #1115, opened 2025-06-21
feat: add tagName parameter to attributeNameCheck

nelstrom:
Allow attributeNameCheck function to receive the tag name as an optional second parameter, enabling more granular control over which attributes are allowed for specific custom elements.

## Summary

This pull request implements an enhancement to the `CUSTOM_ELEMENT_HANDLING.attributeNameCheck` function, allowing it to optionally receive the tag name as a second parameter. This enables developers to create more sophisticated and element-specific attribute validation rules
for custom elements while maintaining full backward compatibility.

## Background & Context

Currently, the attributeNameCheck function in CUSTOM_ELEMENT_HANDLING only receives the attribute name, which limits developers to creating attribute rules that apply globally across all custom elements. This creates a gap when developers need different attributes to be allowed for different custom element types.

For example, a developer might want `<element-one>` to only allow attribute-one and `<element-two>` to only allow attribute-two, but the current API doesn't provide a way to distinguish between these elements during attribute validation.

This enhancement was requested in issue #1093, where the maintainer confirmed they would be open to this feature specifically for custom elements.

## Tasks

- Updated TypeScript interface to make tagName parameter optional in attributeNameCheck
- Modified implementation in src/purify.ts to pass tagName as second parameter
- Added test coverage for the new functionality
- Updated README.md documentation with usage examples

## Dependencies

n/a


cure53:
This looks great, thank you very much! :slightly_smiling_face:

PR #1140, opened 2025-08-29
feat: add from and to for svg animateTransform

alephpiece:
## Summary

This pr adds `from` and `to` for svg animateTransform

## Background & Context

Attributes `from` and `to` are documented in [here](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/animateTransform#attributes) but not in the allow list. This could prevent animations like rotations.

## Tasks

- Add attributes `from` and `to` to svg allow list

## Dependencies

- [x] Resolved dependency
- [ ] Open dependency


cure53:
Sadly, we have to reject this one as those attributes will result in XSS.

Issue #1149, opened 2025-10-03
Update ADD_ATTR to accept a function like CUSTOM_ELEMENT_HANDLING.attributeNameCheck


nelstrom:
Suppose that I want to support the following elements:

1. Valid: `<one attribute-one="1"></one>`
1. Valid: `<two attribute-two="2"></two>`
1. Invalid: `<one attribute-two="2"></one>`
1. Invalid: `<two attribute-one="1"></two>`

I can get close with this:

```js
{
  ADD_TAGS: ['one', 'two'],
  ADD_ATTR: ['attribute-one', 'attribute-two']
}
```

This would allow my valid examples to pass, but it would also allow my invalid examples to pass, because `ADD_ATTR` creates a flat list of allowed attributes that apply to all tags in `ADD_TAGS`.

I propose to allow `ADD_ATTR` to accept a function:

```js
{
  ADD_TAGS: ['one', 'two'],
  ADD_ATTR: (attributeName, tagName) => {
    const allowedAttributes = {
      'one': ['attribute-one'],
      'two': ['attribute-two']
    };
    return allowedAttributes[tagName]?.includes(attributeName) || false;
  }
}
```

This would properly reject the invalid examples while allowing the valid ones.

In #1093, I proposed a solution for this that was specific to custom elements (who's tagname must contain a `-` character). This is a variation on that idea.

cure53:
I think that is a great idea - thanks - and should also no break backwards compatibility - I am all for it 🙂

cure53:
This should be fixed by the latest PR, many thanks again 🙂

PR #1150, opened 2025-10-04
Allow ADD_ATTR and ADD_TAGS to accept a function

nelstrom:
## Summary

This PR addresses issue #1149 by allowing `ADD_ATTR` to accept a function for tag-specific attribute validation.

## Background & Context

Currently, `ADD_ATTR` only accepts a string array, creating a flat list of allowed attributes that applies to all tags. This limitation means you cannot specify that certain attributes should only be allowed on specific tags without writing custom hooks.

## Tasks

**First Commit**: Extends `ADD_ATTR` to accept either a string array (for backwards compatibility) or a function with signature `(attributeName: string, tagName: string) => boolean`. When a function is provided, it enables granular control over which attributes are allowed for specific tags.

**Second Commit**: Extends the same pattern to `ADD_TAGS`, allowing it to accept a function with signature `(tagName: string) => boolean` for consistency. If you prefer to keep the scope limited to the original issue, I'm happy to drop this commit or squash both together.

## Dependencies

n/a


cure53 (approved):

cure53:
Amazing, thank you very much! :heart:

nelstrom:
Thanks for accepting this. I already have a case where I'd like to use this feature. Will you be cutting another release anytime soon?

cure53:
I say thanks :slightly_smiling_face: And yes, maybe this week, latest next week.

PR #1155, opened 2025-10-10
feat: add "mask-type" SVG attribute (resolves #1154)

### LINKED ISSUE #1154: SVG "mask-type" attribute is being removed
### Description
The "[mask-type](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/mask-type)" SVG attribute is not part of the SVG attributes whitelist set and is being removed by default.

#### Input:
```
<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
  <mask id="myMask1" mask-type="alpha">
    <rect
      fill="rgb(10% 10% 10% / 0.4)"
      x="0"
      y="0"
      width="100%"
      height="100%"
    />
    <circle fill="rgb(90% 90% 90% / 0.6)" cx="50" cy="50" r="35" />
  </mask>

  <rect x="0" y="0" width="45" height="45" mask="url(#myMask1)" fill="red" />
</svg>

```
Preview input on [JSFiddle](https://jsfiddle.net/L4k3vjbw/1/).

---

#### Output:

The output was generated using the demo available at [https://cure53.de/purify](https://cure53.de/purify).

```
<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
  <mask id="myMask1">
    <rect
      fill="rgb(10% 10% 10% / 0.4)"
      x="0"
      y="0"
      width="100%"
      height="100%"
    ></rect>
    <circle fill="rgb(90% 90% 90% / 0.6)" cx="50" cy="50" r="35"></circle>
  </mask>

  <rect
    x="0"
    y="0"
    width="45"
    height="45"
    mask="url(#myMask1)"
    fill="red"
  ></rect>
</svg>

```
Preview output on [JSFiddle](https://jsfiddle.net/ujs5dqwe/1/).

---
You can see that the `mask-type` attribute on the `mask` element has been removed.

I did have a look at the attributes set for SVGs in [attrs.ts](https://github.com/cure53/DOMPurify/blob/d996df00a8e326531558bfee609dc7d45a9f00eb/src/attrs.ts#L123) and I can see that the `mask-type` attribute is not in there.

If this wasn't intentional, I can create a PR to add `mask-type` to that set. I wasn't able to find anything in the docs or in any of the closed issues about this but I apologise if this was already discussed in the past.

prasadrajandran:
## Summary

This PR adds the SVG `mask-type` attribute to the default whitelist so that it is not removed during sanitisation.

## Background & Context

The `mask-type` attribute indicates which mask mode, alpha or luminance, to use for the contents of the <mask> element when masking. See [MDN](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/mask-type) for more information about this attribute.

## Tasks

- Added the SVG `mask-type` attribute to the default whitelist
- Added a test case in test/fixtures/expect.mjs that checks if the `mask-type` attribute is not removed during sanitisation
- Ran `npm run test:jsdom`


cure53:
Looks great, thank you very much :)

PR #1168, opened 2025-11-10
Add ADD_FORBID_CONTENTS setting

### LINKED ISSUE #1165: Allow extending, not replacing default "FORBID_CONTENTS" list via setConfig
Hi there!

We'd love to add a couple more elements to the `FORBID_CONTENTS` list. However, when passing a custom array of elements, the default list is no longer used.

There seems to be no way to get the default list either, as there's neither a default export for these, nor can I get the default list via something like DOMPurify.CONFIG....

I rather not want to copy the list into our code by hand, in case a future security updates the defaults, and we're not aware of it.

Thanks!

MariusRumpf:
## Summary

This PR implements the option `ADD_FORBID_CONTENTS` discussed in #1165 and adds two tests to check that.

## Background & Context

DOMPurify has an internal default list for the `FORBID_CONTENTS` setting, which is used when it is not set. To allow extending the default `DEFAULT_FORBID_CONTENTS` this adds a new configuration option `ADD_FORBID_CONTENTS`. This configuration option extends the default or user given list in `FORBID_CONTENTS`.

Given this option, it is no longer required to keep a copy of the `DEFAULT_FORBID_CONTENTS` in code as a user of this library.

Resolves #1165.


cure53:
Whoa, thanks :D

PR #1191, opened 2026-01-16
remove dependency `lodash.sample`

Rotzbua:
## Summary

Use native code instead of dependency `lodash.sample`.

## Background & Context

Dependency can be replaced with 3 lines of code. No need to keep dependency.

## Tasks

- run test


cure53:
Many thanks :slightly_smiling_face:

PR #1210, opened 2026-03-09
fix: engine requirement

### LINKED ISSUE #1209: version 3.3.2 is a breaking change
Hello,

`Version 3.3.2` requires `Node ">= 20"`.
For projects relying on semver ranges this will break their installation if they are on Node < 20 (ie: Node 16.x).

Error:
```
error dompurify@3.3.2: The engine "node" is incompatible with this module. Expected version ">=20". Got "16.15.0"
```

Can we please revert the breaking changes to maintain backwards compatibility in dompurify@3.x versions?

Thanks!

Rotzbua:
## Summary

`engines` requirements causes problems.

## Background & Context

The package is used to distribute finished builds that do not have any direct requirements for the node version used.

However, a modern nodejs environment is required for the build process. A different configuration is provided for this purpose: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#devengines

## Tasks

- CI should go green
- thows warnings on outdated node versions if try to build
- no warning on other installs

## Dependencies

fixes #1209

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,9 +42,6 @@
         "typescript": "^5.6.3",
         "xo": "^0.54.1"
       },
-      "engines": {
-        "node": ">=20"
-      },
       "optionalDependencies": {
         "@types/trusted-types": "^2.0.7"
       }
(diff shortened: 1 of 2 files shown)


cure53:
Thank you very much :bow:

PR #1211, opened 2026-03-11
Getting 3.x branch ready for 3.3.3 release

cure53:
See above


schneidergithub (commented):
The regex update for security feels like there are additional edge cases.

PR #1242, opened 2026-04-13
fix: prevent ADD_ATTR/ADD_TAGS function leaking into subsequent array…

1Jesper1:
When sanitize() is called with ADD_ATTR or ADD_TAGS as a function, the function reference is stored in EXTRA_ELEMENT_HANDLING. A subsequent call that passes ADD_ATTR/ADD_TAGS as an array did not clear the stored function because objectHasOwnProperty(cfg, 'ADD_ATTR') returned true, skipping the conditional reset.

The leaked function is evaluated before URI/tag checks, so a permissive function (returning true) lets dangerous attributes (e.g. javascript: URIs) or forbidden tags (e.g. iframe) through on later calls.

Fix: unconditionally reset tagCheck/attributeCheck to null on every _parseConfig() call, then only set them if the current config provides a function. This ensures no cross-call leakage.

Includes regression tests for both ADD_ATTR and ADD_TAGS leakage scenarios.

## Tasks

Test if the changes are correct and if they don't introduce new problems.

### DIFF
--- a/src/purify.ts
+++ b/src/purify.ts
@@ -635,14 +635,10 @@ function createDOMPurify(window: WindowLike = getGlobal()): DOMPurify {
       }
     }
 
-    /* Prevent function-based ADD_ATTR / ADD_TAGS from leaking across calls */
-    if (!objectHasOwnProperty(cfg, 'ADD_TAGS')) {
-      EXTRA_ELEMENT_HANDLING.tagCheck = null;
-    }
-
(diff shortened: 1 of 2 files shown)


cure53:
This looks great, thank you! Waiting for tests to finish...

cure53:
Many thanks :slightly_smiling_face:

PR #1320, opened 2026-04-29
fix: apply URI validation to attributes allowed via ADD_ATTR callback

### LINKED ISSUE #1319: ADD_ATTR function form bypasses URI validation
In #1149 I proposed allowing `ADD_ATTR` to accept a function, and this was implemented in #1150. I've since discovered that attributes allowed via the function form do not go through URI validation, whereas attributes allowed via the array form (or via `ALLOWED_ATTR`) do.

Consider this input:

```html
<a href="javascript:alert(1)">click</a>
```

**Configuration A (array form):**

```js
DOMPurify.sanitize(input, {
  ADD_ATTR: ['href']
});
```

Result: `<a>click</a>` — the dangerous URI is stripped. This is correct.

**Configuration B (function form):**

```js
DOMPurify.sanitize(input, {
  ADD_ATTR: (attr, tag) => attr === 'href' && tag === 'a'
});
```

Result: `<a href="javascript:alert(1)">click</a>` — the dangerous URI passes through. This is incorrect.

The two configurations should behave identically with respect to URI validation. The function form was intended to enable per-element attribute allowlisting, not to bypass value sanitization.

The root cause is in `_isValidAttribute`: when the `ADD_ATTR` callback returns true, execution falls through to `return true` without checking the value against `IS_ALLOWED_URI`. Attributes allowed via the array form do go through this check.

I'm happy to submit a PR to fix this.

nelstrom:
## Summary

This PR fixes a security flaw where attributes allowed via the function form of `ADD_ATTR` bypassed URI scheme validation, allowing `javascript:` and `data:` URIs to pass through unsanitized.

Fixes #1319

## Background & Context

#1150 added a function form of `ADD_ATTR` with signature `(attributeName, tagName) => boolean` for per-element attribute allowlisting. A design flaw means that when the callback returns `true`, the attribute value is accepted without checking it against `IS_ALLOWED_URI`. This means `javascript:` and `data:` URIs pass through unsanitised, which is inconsistent with the array form of `ADD_ATTR` and the default `ALLOWED_ATTR` set, both of which correctly validate values.

The root cause is that `attributeCheck` was a separate early-exit branch in `_isValidAttribute`, sitting above the URI validation logic in the if-else chain. The fix folds it into the name-permitted check that precedes value validation, so both approval paths reach the same `IS_ALLOWED_URI` checks:

```ts
const nameIsPermitted =
  ALLOWED_ATTR[lcName] ||
  (EXTRA_ELEMENT_HANDLING.attributeCheck instanceof Function &&
    EXTRA_ELEMENT_HANDLING.attributeCheck(lcName, lcTag));

} else if (!nameIsPermitted || FORBID_ATTR[lcName]) {
```

Tasks

- Fix `_isValidAttribute` to route attributeCheck-approved attributes through URI validation
- Add test: javascript: URI is stripped when href is allowed via ADD_ATTR callback
- Add test: safe https: URI is preserved when href is allowed via ADD_ATTR callback

Dependencies

- N/A


x00mario (approved):

nelstrom:
Thanks for accepting this. Any idea when you'll be doing another release?

cure53:
The plan is next week.

PR #1526, opened 2026-07-10
Allow SVG dominant-baseline and text-orientation presentation attributes

### LINKED ISSUE #1525: SVG attribute `dominant-baseline` gets removed
Other properties listed under [W3C - SVG adaptions](https://w3c.github.io/svgwg/svg2-draft/text.html#TextPropertiesAdaptions) might also be missing but I didn't check in detail.
For my specific case only `<text dominant-baseline="middle" />` is required.

Develop-KIM:
Fixes #1525.

`dominant-baseline` was missing from the SVG attribute allow-list, so DOMPurify stripped it (e.g. `<text dominant-baseline="middle">`), even though its baseline siblings `alignment-baseline` and `baseline-shift` are already allowed. This adds it, plus `text-orientation` from the same [W3C SVG2 text property adaptions](https://w3c.github.io/svgwg/svg2-draft/text.html#TextPropertiesAdaptions) group. Both are enum-only presentation attributes with no script or URL surface.

Before / after (default config):

```
IN : <svg><text dominant-baseline="middle">x</text></svg>
OUT: <svg><text>x</text></svg>            // before
OUT: <svg><text dominant-baseline="middle">x</text></svg>   // after
```

You mentioned being happy to add a few safe ones in bulk. I checked the rest of that spec section and deliberately left out `unicode-bidi` (its `bidi-override` value is a text-reordering/spoofing surface, so it felt like your call) and the deprecated `glyph-orientation-horizontal` / `glyph-orientation-vertical`. Happy to add any of those if you'd like.

Added two regression fixtures to `test/fixtures/expect.mjs`. Lint and the jsdom suite pass locally.

Implemented with AI assistance; I've reviewed and verified every change.


x00mario (approved):

cure53:
Excellent, thank you :slightly_smiling_face:

cure53:
The changes look good, thanks @Develop-KIM .

Would you mind adding signatures to your commit? Then we can merge.

Develop-KIM:
Signed the commit, thanks

cure53:
Super, thank you!

Issue #447, opened 2020-06-19
Adding vector-effect to svg whitelist


rohitkatlaa:
> This issue proposes a [bug, feature] which...

### Background & Context
Including an attribute to the svg white-list. 
Attribute: vector-effect

#### Input
```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" width="494" height="367" viewBox="0 0 494 367">
<desc>Created with Fabric.js 3.6.3</desc>
<defs>
</defs>
<rect x="0" y="0" width="100%" height="100%" fill="rgba(0,0,0,0)"/>
<g transform="matrix(3.22 0 0 3.22 270.79 207.79)">
<circle style="stroke: rgb(0,0,0); stroke-width: 9; stroke-dasharray: none; stroke-linecap: butt; stroke-dashoffset: 0; stroke-linejoin: miter; stroke-miterlimit: 4; fill: rgb(237,128,128); fill-opacity: 0; fill-rule: nonzero; opacity: 1;" vector-effect="non-scaling-stroke" cx="0" cy="0" r="30"/>
</g>
</svg>
```

#### Given output
```

<svg viewBox="0 0 494 367" height="367" width="494" version="1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns="http://www.w3.org/2000/svg">
<desc>Created with Fabric.js 3.6.3</desc>
<defs>
</defs>
<rect fill="rgba(0,0,0,0)" height="100%" width="100%" y="0" x="0"></rect>
<g transform="matrix(3.22 0 0 3.22 270.79 207.79)">
<circle r="30" cy="0" cx="0" style="stroke: rgb(0,0,0); stroke-width: 9; stroke-dasharray: none; stroke-linecap: butt; stroke-dashoffset: 0; stroke-linejoin: miter; stroke-miterlimit: 4; fill: rgb(237,128,128); fill-opacity: 0; fill-rule: nonzero; opacity: 1;"></circle>
</g>
</svg>
```

#### Expected output
Output including the vector-effect attribute.

### Feature
As you can see both the svg's are significantly different from eachother.

rohitkatlaa:
Closing this issue because vector-effect can be added to the styles in svg.

PR #1567, opened 2026-08-05
Allow SVG pointer-events and vector-effect presentation attributes

Jaybhade:
## Summary

This pull request adds the `pointer-events` and `vector-effect` SVG presentation attributes to the SVG attribute allow-list. Both are keyword-only presentation attributes, and both were being stripped from otherwise valid SVG.

```js
DOMPurify.sanitize('<svg><rect pointer-events="none" width="10" height="10"/></svg>');
// before: <svg><rect width="10" height="10"></rect></svg>
// after:  <svg><rect pointer-events="none" width="10" height="10"></rect></svg>

DOMPurify.sanitize('<svg><circle vector-effect="non-scaling-stroke" cx="0" cy="0" r="30"/></svg>');
// before: <svg><circle cx="0" cy="0" r="30"></circle></svg>
// after:  <svg><circle vector-effect="non-scaling-stroke" cx="0" cy="0" r="30"></circle></svg>
```

## Background & Context

The SVG list already allow-lists every other keyword-only rendering/interaction presentation attribute in this family — `shape-rendering`, `image-rendering`, `text-rendering`, `color-rendering`, `paint-order`, `visibility`, `display`, `overflow`, `mask-type`, and (since #1526) `dominant-baseline` and `text-orientation`. `pointer-events` and `vector-effect` are the two remaining Baseline-widely-available members of that group that were still missing, so this is a consistency gap rather than a new capability.

The practical symptom is silent visual breakage in generated SVG. `vector-effect="non-scaling-stroke"` is what keeps a stroke width constant under a transform; dropping it makes strokes scale with the shape. This was reported for Fabric.js output in #447 — that issue was closed by its reporter, who moved the property into `style` as a workaround, so the allow-list itself was never revisited. `pointer-events="none"` is the standard way to make an overlay non-interactive and is emitted routinely by charting libraries.

On the security side, neither attribute widens the attack surface:

- **Keyword-only, no URL or script surface.** `pointer-events` takes `bounding-box | visiblePainted | visibleFill | visibleStroke | visible | painted | fill | stroke | all | none`; `vector-effect` takes `none | non-scaling-stroke | non-scaling-size | non-rotation | fixed-position`. Neither accepts a FuncIRI, a URL, or script.
- **Already reachable today.** Both have CSS property counterparts, and `style` is allow-listed, so `style="pointer-events:none"` and `style="vector-effect:non-scaling-stroke"` already survive sanitization unchanged. This repository's own fixtures rely on that — the `#137` SVG image payload in `test/fixtures/expect.mjs` carries `vector-effect: non-scaling-stroke` inside `style`, and the `#33` payload carries `pointer-events:none`. Allowing the presentation-attribute form grants nothing that is not already permitted.
- **No URI-safe shortcut.** Neither attribute is added to the URI-safe set, so values still pass through the normal `IS_ALLOWED_URI` screening rather than bypassing it.
- **Existing controls still apply.** `FORBID_ATTR: ['pointer-events']` still removes it, `USE_PROFILES: { html: true }` still drops the SVG element entirely, and event handlers and `javascript:` URLs on the same element are still stripped.

As with the other entries in `ATTRS.svg`, these names become allow-listed for HTML elements too under the default configuration — the same as `fill`, `display`, and `visibility` today. They remain inert CSS in that position.

## Tasks

- Add `pointer-events` and `vector-effect` to `ATTRS.svg` in `src/attrs.ts`, in their alphabetical positions
- Add two fixtures to `test/fixtures/expect.mjs` covering both attributes, using the payload shape from #447
- Rebuild `dist/` — the bundle diff is exactly the two added strings, with no other build drift
- `npm run lint` (xo) passes clean
- Full jsdom suite passes: 1208 assertions, 0 failures, including the two new fixtures

## Dependencies

- [x] No dependencies on other PRs or API work


x00mario (approved):

cure53:
Heya, lgtm  but we need signed commits :slightly_smiling_face:

Jaybhade:
Done — force-pushed the branch as a single signed commit (`7aa6de3`, verified). The tree is byte-identical to what you reviewed: same 10 files, +20/-8, tree `b7e5acd` either way; it's the same change with x00mario's merge of `main` folded in as the base.

One heads-up: the rewind step auto-closed the PR for a moment and I reopened it, so the approval got dismissed along the way — sorry for the noise. Ready when you are.

cure53:
Looking great, thank you very much :slightly_smiling_face:

x00mario (approved):

PR #1577, opened 2026-08-13
fix: addressing two possible bypasses outside our threat-model

cure53:
fix: addressed an issue with possible bypasses when risky tags are allowed
fix: addressed an issue with possible bypasses when document contexts are mixed


x00mario (approved):

alirezarouhbakhsh:
I found the issue, but I’m not listed as a participant in the PRs.

cure53:
You will be mentioned in the release as stated already via email.

Issue #1118, opened 2025-07-08
CommonJS and ES module


shima2013:
Hi , as i try to upgrade to 3.2.6 , in auto generated purify.cjs.d.ts file i get some content which gives me build error like expected ','
when i check its like bad mix of CommonJS (export =) and ES module (export { ... }) syntax.

the error arises from : 
export { type Config, type DOMPurify, type DocumentFragmentHook, type ElementHook, type HookName, type NodeHook, type RemovedAttribute, type RemovedElement, type UponSanitizeAttributeHook, type UponSanitizeAttributeHookEvent, type UponSanitizeElementHook, type UponSanitizeElementHookEvent, type WindowLike };

// @ts-ignore
export = _default;

cure53:
Oha, thanks for reporting, I am tempted to believe that the commit above already fixes the issue, would you mind checking?

cure53:
Assuming the issue to be sorted, please reopen if not.

shima2013:
still the issue is there in files purify.cjs.d.ts and now in utils.ts

cure53:
Can you be more precise please? What is the issue as of now?

PR #1632, opened 2026-09-22
chore: rolldown migration

ssi02014:
@cure53 @x00mario 

## Summary

Replaces Rollup with [Rolldown](https://rolldown.rs/) as the bundler that produces `dist/`. Build-toolchain only — **no file under `src/` was touched**. The published API, module formats, type declarations and browser support are unchanged.

<br />

## Why

**Written in Rust.** Rolldown is built on [Oxc](https://oxc.rs/), so parsing, transformation and code generation run as native code instead of on the Node.js event loop.

rollup build time
```
> rollup -c config/rollup.config.js

src/purify.ts → ./dist/purify.js, ./dist/purify.min.js, ./dist/purify.es.mjs, ./dist/purify.cjs.js...
created ./dist/purify.js, ./dist/purify.min.js, ./dist/purify.es.mjs, ./dist/purify.cjs.js in 910ms

./dist/types/purify.d.ts → ./dist/purify.es.d.mts...
created ./dist/purify.es.d.mts in 12ms

./dist/types/purify.d.ts → ./dist/purify.cjs.d.ts...
created ./dist/purify.cjs.d.ts in 8ms
```
rolldown build time
```
> rolldown -c config/rolldown.config.js

<DIR>/purify.js.map  asset │ size: 45.72 kB                                                                                                                                                                      
<DIR>/purify.js      chunk │ size: 86.90 kB                                                                                                                                                                      
<DIR>/purify.min.js.map  asset │ size: 38.20 kB                                                                                                                                                                  
<DIR>/purify.min.js      chunk │ size: 28.84 kB                                                                                                                                                                  
<DIR>/purify.es.mjs.map  asset │ size: 45.74 kB                                                                                                                                                                  
<DIR>/purify.es.mjs      chunk │ size: 84.16 kB                                                                                                                                                                  
<DIR>/purify.cjs.js.map  asset │ size: 45.74 kB                                                                                                                                                                  
<DIR>/purify.cjs.js      chunk │ size: 84.17 kB                                                                                                                                                                  
<DIR>/purify.es.d.mts  chunk │ size: 15.66 kB                                                                                                                                                                    
<DIR>/purify.cjs.d.ts  chunk │ size: 15.66 kB                                                                                                                                                                    
                                                                                                                                                                                                                 
✔ rolldown v1.2.9 Finished in 531.78 ms
```

Although `DOMPurify` is a single package, so this isn't a direct measure of perceived performance, it does reduce build time by about `42%`(930ms -> 531.78ms). 

**Plugins are built in.** Four Rollup plugins are now core features, so they are no longer
dependencies: `@rollup/plugin-node-resolve`→ built-in resolution, `@rollup/plugin-replace` → `transform.define`, `@rollup/plugin-terser` → `output.minify`.

**Dependencies:** `rollup`, `rollup-plugin-dts`, the four `@rollup/plugin-*` above and `tslib` are removed; `rolldown` and `rolldown-plugin-dts` are added. Thanks to Rollup API compatibility the config is actually _shorter_ than before (112 → 105 lines; coverage config 67 → 50).

**Babel stays.** `.babelrc` is untouched so the Chrome 51 / Safari 11 down-levelling is bit-for-bit the same pipeline. Only addition: Rolldown strips TypeScript _after_ plugin hooks (the reverse of Rollup), so Babel now needs `@babel/preset-typescript` — the same single-pass arrangement `rollup.coverage.config.js` already used.

<br />

## `dist/` size

|                              |      Rollup |    Rolldown |        Delta |
| ---------------------------- | ----------: | ----------: | -----------: |
| Code (4 bundles + 2 `.d.ts`) |     471,119 |     315,406 | **−155,713** |
| Source maps (4)              |     264,299 |     175,406 |  **−88,893** |
| **Total**                    | **735,418** | **490,812** | **−244,606** |

The published artifacts are **33% smaller overall**. Bundles shrink ~38% because Oxc keeps JSDoc
and legal comments but drops ordinary ones — the code itself is the same size (52,271 → 52,897
bytes stripped of comments and whitespace). 

<br />

## Pre-existing problems fixed along the way

1. **`build:umd` / `build:umd:min` / `build:es` / `build:cjs` never worked** — CLI `-f`/`-o` applied to every config, so all six outputs overwrote one file and `dist/purify.js` ended up containing the CJS _type declaration_ bundle. Now select a single output via `DOMPURIFY_BUILD_TARGET`, each verified byte-identical to the full build.
2. **`BABEL_ENV=rollup` was dead config** — `.babelrc` has no `env` key; builds are hash-identical with it set, bogus, or unset. Removed. `NODE_ENV` retained.

<br />

## Conclusion
This PR updates our build toolchain by replacing Rollup with Rolldown. The goal is to optimize build times and artifact sizes without touching any code under `src/`, maintaining identical runtime behavior and passing existing test suites.


x00mario:
Oh hi, long time no see, thank you, reviewing now :slightly_smiling_face:

x00mario (dismissed):

x00mario:
> Commits must have verified signatures.

Once this has been fixed, we are happy to merge, thanks.

ssi02014:
@x00mario I've updated all the commits with verified signatures as requested. Thanks for reviewing! 😄

scripts/fix-types.js
```diff
@@ -4,14 +4,63 @@ const fs = require('node:fs/promises');
...
+    if (names.length === 0) return statement;
+
+    return [
+      values.length > 0 && `export { ${values.join(', ')} };`,
+      `export type { ${names.join(', ')} };`,
+    ]
+      .filter(Boolean)
+      .join('\n');
+  });
+}
```
ssi02014, line 49: Switching the declaration bundler brings back the TypeScript failure reported in [#1118](https://github.com/cure53/DOMPurify/issues/1118), so this PR
normalises the generated declarations to the shape Rollup used to produce.

#### What changed in the output

`rollup-plugin-dts` and `rolldown-plugin-dts` emit re-exported types differently. Minimal
reproduction — same two input files, only the bundler differs:

```ts
// types.d.ts
export interface Foo {
  a: number;
}
export type Bar = string;

// entry.d.ts
export type { Foo, Bar } from './types';
declare const val: number;
export default val;
```

```ts
// rollup + rollup-plugin-dts
export { val as default };
export type { Bar, Foo };

// rolldown + rolldown-plugin-dts
export { type Bar, type Foo, val as default };
```

Nothing in `src/` or in the `tsc` output changed — `dist/types/*.d.ts` still contains
`export type { Config } from './config'`. Only the declaration bundler's code generation differs.

#### Why that matters

Inline `type` specifiers inside `export { ... }` require **TypeScript 4.5**. 

| TypeScript | before this change            | after |
| ---------- | ----------------------------- | ----- |
| 4.4.4      | `error TS1005: ',' expected.` | ok    |
| 4.5.5      | ok                            | ok    |
| 5.6.3      | ok                            | ok    |

That error is exactly what #1118 reported

#### The fix

`scripts/fix-types.js` now rewrites those statements back to the split form before the existing
CommonJS post-processing:

```ts
export { type Config, ..., _default as default };
// becomes
export { _default as default };
export type { Config, ... };
```

The type surface is identical and both declaration files now match what `main` branch

x00mario (approved):

